PR #39, opened 2023-04-12
feat(env): read variables from .env file

gauravsaini:
read vars from env file 
otherwise use the fallback

### DIFF
--- a/src/utils/chain.ts
+++ b/src/utils/chain.ts
@@ -6,9 +6,9 @@ export const createModel = (customApiKey: string) =>
   new OpenAI({
     openAIApiKey:
       customApiKey === "" ? process.env.OPENAI_API_KEY : customApiKey,
-      temperature: process.env.TEMPRATURE  || 0.9,
-      modelName: process.env.MODEL || "gpt-3.5-turbo",
-      maxTokens: process.env.MAX_TOKENS || 300,
+      temperature: process.env.OPENAI_TEMPRATURE  || 0.9,
+      modelName: process.env.OPENAI_MODEL || "gpt-3.5-turbo",
+      maxTokens: process.env.OPENAI_MAX_TOKENS || 300,
   });
 
 const startGoalPrompt = new PromptTemplate({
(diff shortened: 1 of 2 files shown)


src/utils/chain.ts
```diff
@@ -6,9 +6,9 @@ export const createModel = (customApiKey: string) =>
   new OpenAI({
     openAIApiKey:
       customApiKey === "" ? process.env.OPENAI_API_KEY : customApiKey,
-    temperature: 0.9,
-    modelName: "gpt-3.5-turbo",
-    maxTokens: 300,
+      temperature: process.env.TEMPRATURE  || 0.9,
```
awtkns, line 9: @gauravsaini thoughts on prefixing this with OPENAI_

gauravsaini: These `in future` can be with anything (not necessarily OPENAI )
I think AGENTGPT_ would be a better prefix 
Thoughts ?

awtkns: @gauravsaini my thoughts with the prefix are that in the future we will most likely have multiple model that will be running at once in which case we would need to distinguish between tempratures

awtkns (changes requested):

mrayonnaise:
I have been working on something similar to this too. Maybe we can combine efforts? I am adding the MAX_LOOPS as a config item for phase 1 and then adding it as a setting in the UI in phase 2. My work in progress (WIP) is here:
https://github.com/reworkd/AgentGPT/pull/51

gauravsaini:
@awtkns  : Done

src/utils/chain.ts
```diff
@@ -6,9 +6,9 @@ export const createModel = (customApiKey: string) =>
   new OpenAI({
     openAIApiKey:
       customApiKey === "" ? process.env.OPENAI_API_KEY : customApiKey,
-    temperature: 0.9,
-    modelName: "gpt-3.5-turbo",
-    maxTokens: 300,
+      temperature: process.env.OPENAI_TEMPRATURE  || 0.9,
```
chozzz, line 9: Is it a typo? `OPENAI_TEMPRATURE`?

PR #52, opened 2023-04-12
chore: show "Copied!" on copy click

hahayusuf:
Created a function that edits the text of the "copy" button to "copied!" when it is clicked.


hahayusuf:
That's a good idea, thanks.

Will implement it.

asim-shrestha:
Hey @hahayusuf great work! We should however avoid using document.getElementById when we can and use react standard ways to implement this. See the copy button used in message hover as an example

hahayusuf:
I have updated it to add the timeout, and react referencing instead of elementId.

I appreciate your feedback. How is it after the last commit?

hahayusuf:
Addressed by #84

PR #76, opened 2023-04-13
:lock: Authentication and User Accounts

awtkns:


src/server/auth.ts
```diff
@@ -28,6 +31,19 @@ declare module "next-auth" {
   // }
 }
 
+const providers = [
+  GoogleProvider({
+    clientId: env.GOOGLE_CLIENT_ID,
+    clientSecret: env.GOOGLE_CLIENT_SECRET,
+    allowDangerousEmailAccountLinking: true,
```
asim-shrestha, line 38: Is this fine to leave as true? Will we be updating once we get a prod ready release

src/server/api/routers/agentRouter.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const agentRouter = createTRPCRouter({
+  create: protectedProcedure
+    .input(z.object({ name: z.string(), goal: z.string() }))
+    .mutation(async ({ input, ctx }) => {
+      const agent = await prisma.agent.create({
+        data: {
+          name: input.name,
+          goal: input.goal,
+          userId: ctx.session?.user?.id,
```
asim-shrestha, line 14: Probably change to NoSQL in the future?

asim-shrestha (approved):

src/server/auth.ts
awtkns: This allow users to login with all providers and have them be be the same account. This is safe as long as the providers are well known and secure.

hrada777:
Z

PR #84, opened 2023-04-13
Enhancement: Option to change OpenAI API Endpoint

hahayusuf:
You can now set a custom API base URL through the env file using the `API_BASE_URL` variable. When kept blank (""), it will use the default base URL as per the previous implementation.

The README, and setup scripts were updated to incorporate the new variable.

### DIFF
--- a/src/components/ChatWindow.tsx
+++ b/src/components/ChatWindow.tsx
@@ -21,6 +21,7 @@ interface ChatWindowProps {
 }
 
 const messageListId = "chat-window-message-list";
+const copyButtonId="btn-copy-01";
 
 const ChatWindow = ({ messages, children, className }: ChatWindowProps) => {
   const [hasUserScrolled, setHasUserScrolled] = useState(false);
@@ -119,16 +120,29 @@ const MacWindowHeader = () => {
       .catch(console.error);
   };
 
-  const copyElementText = (elementId: string) => {
+  const changeInnerText = (elementId: string, text: string) => {
     const element = document.getElementById(elementId);
     if (!element) {
       return;
     }
 
+    element.innerText="Copied!"
+
+  }
+
+  const copyElementText = (elementId: string, buttonId: string) => {
+    const element = document.getElementById(elementId);
+    if (!element) {
+      return;
+    }
+    
     const text = element.innerText;
     navigator.clipboard.writeText(text).then(
       () => {
         console.info("Copied text to clipboard");
+        if(buttonId){
+          changeInnerText(buttonId, "Copied")
+        }
       },
       () => {
         console.error("Failed to copy text to clipboard");
@@ -157,10 +171,10 @@ const MacWindowHeader = () => {
       </div>
       <div
         className="mr-1 flex cursor-pointer items-center gap-2 rounded-full border-2 border-white/30 p-1 px-2 hover:bg-white/10"
-        onClick={(): void => copyElementText(messageListId)}
+        onClick={(): void => copyElementText(messageListId, copyButtonId)}
       >
         <FaClipboard size={12} />
-        <p className="font-mono">Copy</p>
+        <p id="btn-copy-01" className="font-mono">Copy</p>
       </div>
     </div>
   );
@@ -172,6 +186,7 @@ const ChatMessage = ({ message }: { message: Message }) => {
   const handleCopyClick = () => {
     void navigator.clipboard.writeText(message.value);
     setCopied(true);
+
   };
   useEffect(() => {
     let timeoutId: NodeJS.Timeout;
--- a/src/env/schema.mjs
+++ b/src/env/schema.mjs
@@ -20,7 +20,8 @@ export const serverSchema = z.object({
     // VERCEL_URL doesn't include `https` so it cant be validated as a URL
     process.env.VERCEL ? z.string() : z.string().url(),
   ),
-  OPENAI_API_KEY: z.string()
+  OPENAI_API_KEY: z.string(),
+  API_BASE_URL: z.string()
 });
 
 /**
@@ -34,6 +35,7 @@ export const serverEnv = {
   NEXTAUTH_SECRET: process.env.NEXTAUTH_SECRET,
   NEXTAUTH_URL: process.env.NEXTAUTH_URL,
   OPENAI_API_KEY: process.env.OPENAI_API_KEY,
+  API_BASE_URL: process.env.API_BASE_URL
 };
 
 /**
--- a/src/utils/chain.ts
+++ b/src/utils/chain.ts
@@ -2,14 +2,31 @@ import { OpenAI } from "langchain/llms/openai";
 import { PromptTemplate } from "langchain/prompts";
 import { LLMChain } from "langchain/chains";
 
-export const createModel = (customApiKey: string) =>
-  new OpenAI({
+export const createModel = (customApiKey: string) => {
+
+  const customBaseUrl=process.env.API_BASE_URL;
+  
+  if(customBaseUrl=="" || customBaseUrl==undefined){
+    return new OpenAI({
+      openAIApiKey:
+        customApiKey === "" ? process.env.OPENAI_API_KEY : customApiKey,
+      temperature: 0.9,
+      modelName: "gpt-3.5-turbo",
+      maxTokens: 300,
+    })
+  }
+
+  return new OpenAI({
     openAIApiKey:
       customApiKey === "" ? process.env.OPENAI_API_KEY : customApiKey,
     temperature: 0.9,
     modelName: "gpt-3.5-turbo",
     maxTokens: 300,
-  });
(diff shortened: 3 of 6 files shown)


pengmaochang:
@hahayusuf the file "setup.sh" write like this may be better
```
#!/bin/bash
cd "$(dirname "$0")" || exit

echo -n "Enter your OpenAI Key (eg: sk...): "
read OPENAI_API_KEY
echo -n "Enter your API_BASE_URL IS YOU NEEDED ( press enter for default): "
read API_BASE_URL

NEXTAUTH_SECRET=$(openssl rand -base64 32)

ENV="NODE_ENV=development\n\
NEXTAUTH_SECRET=$NEXTAUTH_SECRET\n\
NEXTAUTH_URL=http://localhost:3000\n\
OPENAI_API_KEY=$OPENAI_API_KEY\n\
DATABASE_URL=file:../db/db.sqlite\n
API_BASE_URL=$API_BASE_URL\n"
...
```

src/env/schema.mjs
```diff
@@ -20,7 +20,8 @@ export const serverSchema = z.object({
     // VERCEL_URL doesn't include `https` so it cant be validated as a URL
     process.env.VERCEL ? z.string() : z.string().url(),
   ),
-  OPENAI_API_KEY: z.string()
+  OPENAI_API_KEY: z.string(),
+  API_BASE_URL: z.string()
```
awtkns, line 24: Might be best to prefix this with OPENAI_ as we are doing in the field above :)

Mohamedhoussein:
I am interested

src/env/schema.mjs
hahayusuf: I have resolved this with the last commit.

src/components/ChatWindow.tsx
```diff
@@ -123,16 +124,29 @@ const MacWindowHeader = () => {
...
   };
 
-  const copyElementText = (elementId: string) => {
+  const changeInnerText = (elementId: string, text: string) => {
     const element = document.getElementById(elementId);
     if (!element) {
       return;
     }
 
+    element.innerText="Copied!"
```
Jshen123, line 133: Hi @hahayusuf ,  change this to "text" param. This is hardcoded right now.

src/components/ChatWindow.tsx
```diff
@@ -161,10 +175,10 @@ const MacWindowHeader = () => {
       </div>
       <div
         className="mr-1 flex cursor-pointer items-center gap-2 rounded-full border-2 border-white/30 p-1 px-2 hover:bg-white/10"
-        onClick={(): void => copyElementText(messageListId)}
+        onClick={(): void => copyElementText(messageListId, copyButtonId)}
       >
         <FaClipboard size={12} />
-        <p className="font-mono">Copy</p>
+        <p id="btn-copy-01" className="font-mono">Copy</p>
```
Jshen123, line 181: following the [BEM](https://getbem.com/naming/) convention, can you change this to something like "copy--button"

src/utils/chain.ts
```diff
@@ -2,14 +2,31 @@ import { OpenAI } from "langchain/llms/openai";
...
       customApiKey === "" ? process.env.OPENAI_API_KEY : customApiKey,
     temperature: 0.9,
     modelName: "gpt-3.5-turbo",
     maxTokens: 300,
-  });
+  },{
+    basePath : baseUrl
+  })
+
+};
```
Jshen123, line 29: Hi @hahayusuf, you can probably refactor this to something like this:

   ```
 return new OpenAI({
      openAIApiKey:
        customApiKey === "" ? process.env.OPENAI_API_KEY : customApiKey,
      temperature: 0.9,
      modelName: "gpt-3.5-turbo",
      maxTokens: 300,
    }, {
    basePath : process.env.OPENAI_API_BASE_URL || undefined
  })
```

**Review by Jshen123 — CHANGES_REQUESTED** [review]
Hi @hahayusuf, thanks for making this PR :) this is a very useful change

I have included some feedback. Also, there will likely be merge conflicts since a lot has changed since this PR was made

**Comment by jondwillis** [comment]
This is also useful for analytics and caching layers such as https://helicone.ai

**Comment by hahayusuf** [comment]
> Hi @hahayusuf, thanks for making this PR :) this is a very useful change
> 
> 
> 
> I have included some feedback. Also, there will likely be merge conflicts since a lot has changed since this PR was made


Thanks for your feedback! 

Will let you know when it's ready to merge.

**Comment by hahayusuf** [comment]
@Jshen123  

The merge conflicts have now been resolved and can be merged automatically.

I have refactored this to use the `ModelSettings` interface for setting the `customBaseUrl`.

- Input field created for customBaseUrl under advanced settings in settings dialog.
- Fixed other comments related to "Copy!" message when clicking the copy button.

**Review by Jshen123 — CHANGES_REQUESTED** [review]
Hi @hahayusuf  loooks really good! 🎉  I made some minor tweaks here. (you can see the changes once you pull into your local branch)
https://github.com/reworkd/AgentGPT/pull/246

I highlighted the main ones. everything else is just automatic syntax formatting from linter

**Comment by hahayusuf** [comment]
@Jshen123 

Thanks for your feedback. 

I have added your "minor tweaks" commit. 

Then I noticed some merge conflicts, which I have now resolved.

**Review by Jshen123 — APPROVED** [review]
approved!

**Comment by awtkns** [comment]
I will take a look at this shortly!

File: src/utils/prompts.ts [hunk 5]
```diff
@@ -13,6 +13,12 @@ export const createModel = (settings: ModelSettings) =>
     modelName:
       settings.customModelName === "" ? GPT_35_TURBO : settings.customModelName,
     maxTokens: 400,
+  },
+  {
+    basePath:
+      settings.customBaseUrl === ""
+        ? process.env.OPENAI_API_BASE_URL
```
awtkns, line 20:
```
env. OPENAI_API_BASE_URL
```

awtkns, line 20:
```suggestion
      settings.customBaseUrl === ""
        ? env.OPENAI_API_BASE_URL
```
Above we will need to import the client schema from `env/client.mjs`. Doing so allows us to preparse and validate the environment variables. More info can be found here: https://create.t3.gg/en/usage/env-variables

awtkns (changes requested):
I have a slight security concern about allowing users to change the base url in our production environments. Doing so could potentially expose our keys if we allow them to change the baseurl + not have to input their own key. Ideally we will probably want to keep this client side if possible. Thoughts @asim-shrestha?

hahayusuf:
> I have a slight security concern about allowing users to change the base url in our production environments. Doing so could potentially expose our keys if we allow them to change the baseurl + not have to input their own key. Ideally we will probably want to keep this client side if possible. Thoughts @asim-shrestha?

It is a valid concern @awtkns. We can protect against it by implementing two safeguards:
-Frontend: hiding the custom url input unless a `customApiKey` is set.
-Backend: when `settings.customBaseUrl` is set, `env.OPENAI_API_KEY` cannot be used, only `customApiKey`. A simple guard clause should do the trick.

Jshen123:
> > I have a slight security concern about allowing users to change the base url in our production environments. Doing so could potentially expose our keys if we allow them to change the baseurl + not have to input their own key. Ideally we will probably want to keep this client side if possible. Thoughts @asim-shrestha?
> 
> It is a valid concern @awtkns. We can protect against it by implementing two safeguards: -Frontend: hiding the custom url input unless a `customApiKey` is set. -Backend: when `settings.customBaseUrl` is set, `env.OPENAI_API_KEY` cannot be used, only `customApiKey`. A simple guard clause should do the trick.

@awtkns that's a really good callout. @hahayusuf  agree with your suggested solution (only change here would be to use either disable or disable + hide)

@awtkns @asim-shrestha, if we are not comfortable with custom URL in production, perhaps we can rollback URL input feature and release the rest of the code (so users can at least use their own URLs locally). Thoughts?

hahayusuf:
Update: I have implemented both safeguards discussed. 

Thanks for your feedback @Jshen123 

Look forward to hearing what others think.

asim-shrestha:
Hey sorry @hahayusuf, this has flown under the radar. Unfortunate because it seems we have a bunch more conflicts. I can have a look into manually fixing this stuff if you're not able. 

Configuring endpoint locally via ENV seems like a good win. For endpoint in settings, given the safe guards it should be ok but maybe some unit tests are in order to validate this. Otherwise, perhaps we can provide a dropdown of options people can use. Not sure what other endpoints they might try.

hahayusuf:
Hey @asim-shrestha 

Thanks for offering to resolve some of the conflicts.

I agree, using env to set a custom base url should be a good enough win.

Will work on that. May need help ironing out conflicts.

asim-shrestha:
Awesome!

awtkns:
Closed as this has been already implemented

Issue #33, opened 2023-04-12
🤖 Add option to use GPT-4


avelican:
Many of your users have access to GPT-4 already. I checked the settings but there is no option to use it. I checked the code, and the model name is baked-in as gpt-3.5-turbo. It would be very nice to have a GPT-4 option considering the difference in quality.
(Might be worth adding a notice that using GPT-4 will be slower and more expensive.)

OVAWARE:
I would agree. Allow users to choose between GPT-3.5, 3 and 4 when they have there custom token.

gauravsaini:
https://github.com/reworkd/AgentGPT/pull/39

asim-shrestha:
Thanks @gauravsaini that will help on the dev side, for production usage @Jshen123 is working on something

Jshen123:
PR for #88

yourepicservices:
Model input has been added in settings

PR #88, opened 2023-04-13
feat(#33): implement model input

Jshen123:
**Description:**
allow user to input custom model name

image

### DIFF
--- a/src/components/SettingsDialog.tsx
+++ b/src/components/SettingsDialog.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import Button from "./Button";
-import { FaKey } from "react-icons/fa";
+import { FaKey, FaMicrochip } from "react-icons/fa";
 import Dialog from "./Dialog";
 import Input from "./Input";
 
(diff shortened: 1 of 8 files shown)


src/components/SettingsDialog.tsx
```diff
@@ -56,11 +60,23 @@ export default function SettingsDialog({
             link.
           </a>
         </p>
+        <Input
```
awtkns, line 63: Ideally I think this should be a drop down. But we can make that improvement later.

src/components/SettingsDialog.tsx
```diff
@@ -56,11 +60,23 @@ export default function SettingsDialog({
...
+          left={
+            <>
+              <FaMicrochip />
+              <span className="ml-2">Model:</span>
+            </>
+          }
+          placeholder={""}
```
awtkns, line 70: Might be good to have a place holder here. A default value would be even better 😊

src/components/SettingsDialog.tsx
asim-shrestha: Told him drop down could wait for a diff push

src/components/SettingsDialog.tsx
asim-shrestha: lets do a placeholder of gpt-3.5-turbo

asim-shrestha (approved):

PR #110, opened 2023-04-14
📃 Window for current tasks

asim-shrestha:


awtkns (approved):
Looks like there are some alignment issues but this is already a big improvement!

awtkns:
@asim-shrestha we should probably be removing (or marking) tasks that are completed, and also scrolling to the bottom.

PR #112, opened 2023-04-14
feat(#88): add temperature input

Jshen123:
**Description:**
1. added Accordion Component
image

2. Users can now customize temperature (you can find it under advanced settings)
image

3. Added tooltip Component 
a. available as standalone component and can be enabled as part of label component.
image


awtkns:
It might be good to explain what temperature is in the future :) We can use a hover component from radix-ui

PR #115, opened 2023-04-14
feat(#33): add model name editable

Jshen123:
Description:
Editable Combobox for model name
image


awtkns:
This pr also seems to have the changes from #112

awtkns:
Maybe we could drop that commit?

Jshen123:
> Maybe we could drop that commit?

Thanks for catching this! :) Reverted ✅

PR #134, opened 2023-04-14
Update .env.example

Mikasaak:


awtkns:
wont fix this, we want users to know that they need to be changed

Issue #122, opened 2023-04-14
Website needs a bit of work...


Tristan3D3:
On my laptop, the website is kinda broken with long ugly scroll bars and text clipping

This might not be a high priority, but if someone has the time, they should fix it. 

![Css_Issue](https://user-images.githubusercontent.com/116461571/232056788-bff08de4-7404-470e-8339-31cba24f1e03.png)

rahul-ghimire-au6:
hey @Tristan3D3 hope you are doing great! 
can you please share some more info on this issue? maybe sharing your laptop's screen resolution might help to pin-point this issue so that I can re-create this issue in my laptop I tried to re-create this issue by checking in different screen resolution in chrome dev mode but didn't faced it on my end 
common screen resolution I tested: 
small:- 1366 * 768
medium:- 1536 * 864
large:- 1920 * 1080

Tristan3D3:
> hey @Tristan3D3 hope you are doing great! can you please share some more info on this issue? maybe sharing your laptop's screen resolution might help to pin-point this issue so that I can re-create this issue in my laptop I tried to re-create this issue by checking in different screen resolution in chrome dev mode but didn't faced it on my end common screen resolution I tested: small:- 1366 * 768 medium:- 1536 * 864 large:- 1920 * 1080

Here's my screen specifications :
- Size : 15.6"
- Resolution 1336x768

I think it might help to know that the Chrome tab bar might be affecting the page resolution because the issue is resolved if I go full screen (F11).

Tristan3D3:
> > hey @Tristan3D3 hope you are doing great! can you please share some more info on this issue? maybe sharing your laptop's screen resolution might help to pin-point this issue so that I can re-create this issue in my laptop I tried to re-create this issue by checking in different screen resolution in chrome dev mode but didn't faced it on my end common screen resolution I tested: small:- 1366 * 768 medium:- 1536 * 864 large:- 1920 * 1080
> 
> Here's my screen specifications :
> 
> * Size : 15.6"
> * Resolution 1336x768
> 
> I think it might help to know that the Chrome tab bar might be affecting the page resolution because the issue is resolved if I go full screen (F11).

Update : 

Recreated the issue in chrome dev mode :

Resolution : 1366x600
Scale : 50%

![image](https://user-images.githubusercontent.com/116461571/232075337-c7679019-6032-4ec7-9d5b-d47afbf40215.png)

awtkns:
[...]

PR #136, opened 2023-04-14
Resetting the screen height to auto.

Tristan3D3:
Resetting the screen height to auto because it caused some funny glitches. #22 & #122 

This should hopefully fix it :)

PR #159, opened 2023-04-15
Tailwind css fix for issue #122

Tristan3D3:
This is just a follow-up to PR #136

It fixes issue #122 
All changes are done in tailwind CSS

Hope this fixes the problems :)


asim-shrestha:
Unfortunately this causes issues with larger devices as we see a gap in the bottom :(

image

Tristan3D3:
> Unfortunately this causes issues with larger devices as we see a gap in the bottom :(


I see.

Trying to fix the issue

Issue #173, opened 2023-04-16
Display issue when there's a lot of tasks listed


JohnPevien:
When an excessive number of tasks are added, the sidebar that displays the task list expands the container, causing it to become extremely long. This also results in the addition of a scroll and prevents the process window from being displayed properly.

![agent-gpt-ui-issue](https://user-images.githubusercontent.com/36871970/232280958-2060ada3-791e-4a1f-84da-6474fa84cd74.gif)
![display issue when there's a lot of tasks added](https://user-images.githubusercontent.com/36871970/232280959-c5988923-e173-48a0-93ed-9c55551be6b5.png)

JohnPevien:
Submitted a PR #174

PR #174, opened 2023-04-16
Fix display issue when there's a lot of tasks listed #173

JohnPevien:
issue related to #173 

- add max-height
- add there's a bit of space between the last task container and the parent container


awtkns:
Hi @JohnPevien thanks for this! However I believe that this is fixed in #180

PR #186, opened 2023-04-16
fix: remove rogue zero

ka-fuachie:
Same fix as in #138


asim-shrestha:
Ah, just pushed this myself before seeing this, sorry!

ka-fuachie:
😂😂 No worries

PR #219, opened 2023-04-18
Key validation

matthope1:
Added Key validation to the setup script.

The keys from open ai have 51 characters in total and are prefixed with "sk-".
So this regex captures that case ^sk-[a-zA-Z0-9]{48}$

I've also added the same check within the settingsDialog component & It won't save an API key unless it has a valid format.
It will also notify the user with an alert.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,8 +1,21 @@
 #!/bin/bash
 cd "$(dirname "$0")" || exit
 
+is_valid_key() {
+  local api_key=$1
+  local pattern="^sk-[a-zA-Z0-9]{48}$"
+  [[ $api_key =~ $pattern ]] && return 0 || return 1
+}
+
 echo -n "Enter your OpenAI Key (eg: sk...): "
 read OPENAI_API_KEY
+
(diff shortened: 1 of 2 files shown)


jmrada14:
The RegExp used `/^sk-[a-zA-Z0-9]{48}$/` is returning no match for valid OpenAI API keys.

Valid keys that are failing:
sk-proj-pF9cTsH5RrqQhVQK5kfiT3BlbkFJ3ISRnmVg8d3DKcqL6aPa
sk-proj-40qHJ5ChqF12c1A9l8VrT3BlbkFJt5XkT7qspHLhwPIJk5Ue
sk-proj-nNBQEiRMNrtWhOx1sPxGT3BlbkFJ6GseFSbmeVDVcey0I0GI


"The keys from open ai have 51 characters in total" not sure this is true anymore, all new keys I get are 56 characters in total. but that might change in the future. 

I have created a PR to use `/^sk-[a-zA-Z0-9]/` which does not validate the length.  
https://github.com/reworkd/AgentGPT/pull/1546

PR #242, opened 2023-04-18
feat: [#232] Added a bigger UI for Goal input

### LINKED ISSUE #232: Bigger Chat UI
Currently the chat (more obvious on mobile) takes place in a little scrollable window which reminds me of the hood old tamagotchi times. Anything against of changing that to a more practical layout? I would make a PR for it but wanted to check in before hand

giri-madhan:
### **Scope of change**
- Added a bigger UI for `Goal` input using the `textarea` HTML element.
- As a `textarea` is used, users can now enter goals in multiple lines by pressing the `Shift + Enter` keys to create a new line.

### **Key Files**
- Input.tsx
- Label.tsx

### Screenshots
**Larger Screens**
Screenshot 2023-04-19 at 2 06 55 AM

**Mobile**
Screenshot 2023-04-19 at 2 06 29 AM

### **Notes**
This fixes #232

### DIFF
--- a/src/components/Label.tsx
+++ b/src/components/Label.tsx
@@ -9,13 +9,18 @@ interface LabelProps {
 }
 
 const Label = ({ type, left, toolTipProperties }: LabelProps) => {
+  const isTypeTextArea = () => {
+    return type === "textarea";
+  };
+  
   return (
     <Tooltip
       child={
         <div
           className={`center flex items-center rounded-xl rounded-r-none ${
             type !== "range" ? "border-r-0 border-white/10 md:border-[2px]" : ""
-          }  py-2 text-sm font-semibold tracking-wider transition-all sm:py-3 md:pl-3 md:text-lg`}
+          }  py-2 text-sm font-semibold tracking-wider transition-all sm:py-3 md:pl-3 md:text-lg 
+          ${isTypeTextArea() && 'md:h-20'}`}
         >
           {left}
         </div>
--- a/src/components/Input.tsx
+++ b/src/components/Input.tsx
@@ -66,7 +66,7 @@ const Input = (props: InputProps) => {
     inputElement = (
       <textarea
         className={clsx(
-          "border:black delay-50 w-full h-20 rounded-xl bg-[#3a3a3a] text-sm tracking-wider outline-0 transition-all placeholder:text-white/20 hover:border-[#1E88E5]/40 focus:border-[#1E88E5] py-3 md:text-lg border-[2px] border-white/10 px-2",
+          "border:black delay-50 w-full resize-none h-20 rounded-xl bg-[#3a3a3a] text-sm tracking-wider outline-0 transition-all placeholder:text-white/20 hover:border-[#1E88E5]/40 focus:border-[#1E88E5] py-3 md:text-lg border-[2px] border-white/10 px-2",
           disabled && " cursor-not-allowed hover:border-white/10",
           left && "md:rounded-l-none"
         )}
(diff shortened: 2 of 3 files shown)


giri-madhan:
Hello!

I would appreciate it if you could review the implementation and let me know if it is satisfactory, or if any changes need to be made.

Thank you.

src/components/Label.tsx
```diff
@@ -9,13 +9,18 @@ interface LabelProps {
...
+  
   return (
     <Tooltip
       child={
         <div
           className={`center flex items-center rounded-xl rounded-r-none ${
             type !== "range" ? "border-r-0 border-white/10 md:border-[2px]" : ""
-          }  py-2 text-sm font-semibold tracking-wider transition-all sm:py-3 md:pl-3 md:text-lg`}
+          }  py-2 text-sm font-semibold tracking-wider transition-all sm:py-3 md:pl-3 md:text-lg 
+          ${isTypeTextArea() && 'md:h-20'}`}
```
appinteractive, line 23: Is a fixed height a good idea? Should this not be more like a min height instead? 

I don't know currently if the field is resizing on input.

giri-madhan: We have a separate `Label` component whose height also needs to be updated concurrently with the height of the field when new lines are added. However, there is an delay in this solution and feels _hacky_ as seen in this video. 

One potential solution could be to set a fixed height for the `Goal` field for now (as in the current version of the PR), and treat the auto-resize functionality as an enhancement to be addressed later on. This would allow us to have a more stable and predictable behavior for the Label component in the meantime. What you think about this @appinteractive ?

https://user-images.githubusercontent.com/87971509/232950607-97cc5652-bdc9-422c-bd85-2a774a36a18f.mov

giri-madhan: This is how the current implementation would look after the changes in this Pull Request

https://user-images.githubusercontent.com/87971509/232951645-4c7d0008-7f2d-489d-8af6-510f19a6f99d.mov

asim-shrestha:
Sweet! Some minor notes:

image

For whatever reason the scroll bar styling (On desktop) isn't consistent with the global styling the apps defined. Also, when manually enlarging the window, the border on the left of the input doesn't also grow

giri-madhan:
> Sweet! Some minor notes:
> 
> image
> 
> For whatever reason the scroll bar styling (On desktop) isn't consistent with the global styling the apps defined. Also, when manually enlarging the window, the border on the left of the input doesn't also grow

@asim-shrestha 

1) I suggest disabling manual window enlargement as it interferes with the rest of the styling. Let me make this change. 
2) As for the scroll bar styling, I was unable to replicate the issue in Chrome and Safari on Mac using the vercel [link](https://agent-gpt-git-fork-giri-madhan-feat-232-bigger-chat-ui-reworkd.vercel.app/). Could you please provide more information on the browser or operating system where this bug is occurring? This will help better understand the issue.

Thank you for you reply @asim-shrestha !

src/components/Label.tsx
appinteractive: Why is the left lable delayed in size? Is there any JavaScript in setting a height? Should work if its just `h-full` or do I miss something?

michael-pfister:
this way we will never get an 'ok' and 'ready to review' checkmark

src/components/Label.tsx
giri-madhan: Hi @appinteractive,

Thank you for your comment! I appreciate your suggestion of using h-full, but unfortunately, it did not result in the label container automatically scaling to match its parent. It seems like the Input and Label are independent components, which could be causing the issue. I'm not entirely certain about the root cause.

If you are interested, it would be great if you could take a look into this issue. Thank you again for the suggestion!

PR #284, opened 2023-04-20
Update

Cs4K1Sr4C:
-🌐 Added translations
-🌟 Improved the quality of already existing prompts. Let's make the experience more inclusive!
-🆕 Introducing two new prompts:
    1. "Agent Prompt Creator" prompt for generate prompt which creates an Agent
    2. "Agent Creator" prompt based on its build-up prompt
-🆕 Added an input field to change the Agent's language. Let's make it easier to use in different regions!

### DIFF
--- a/src/components/ChatWindow.tsx
+++ b/src/components/ChatWindow.tsx
@@ -178,9 +178,12 @@ const MacWindowHeader = (props: HeaderProps) => {
       <PopIn delay={0.6}>
         <div className="h-3 w-3 rounded-full bg-green-500" />
       </PopIn>
-      <div className="flex flex-grow font-mono text-sm font-bold text-gray-600 sm:ml-2">
+      <Expand
+        delay={1}
+        className="flex flex-grow font-mono text-sm font-bold text-gray-600 sm:ml-2 "
+      >
         {props.title}
-      </div>
+      </Expand>
       {props.onSave && (
         <WindowButton
           delay={0.8}


public/locales/$LOCALE/$NAMESPACE.json/public/locales/ch/translation.json
```diff
@@ -0,0 +1,53 @@
+{
```
awtkns, line 1: Seems like folder `public/locales/$LOCALE/$NAMESPACE` has been added to git. Was this intentional? I noticed you added this to the git ignore.

src/utils/prompts.ts
```diff
@@ -17,18 +17,30 @@ export const createModel = (settings: ModelSettings) =>
...
   template:
-    "You are an AI task creation agent. You have the following objective `{goal}`. You have the following incomplete tasks `{tasks}` and have just executed the following task `{lastTask}` and received the following result `{result}`. Based on this, create a new task to be completed by your AI system ONLY IF NEEDED such that your goal is more closely reached or completely reached. Return the response as an array of strings that can be used in JSON.parse() and NOTHING ELSE",
-  inputVariables: ["goal", "tasks", "lastTask", "result"],
+    "You are You are an AI task creation agent and your objective is: `{goal}`. You have the following incomplete tasks `{tasks}` and have just executed the following task `{lastTask}` and received the following solution `{result}`. Based on this, create a new task to be completed by your AI system ONLY IF NEEDED such that your goal is more closely reached or completely reached. Use the `{customLanguage}` language to create the new task. Respond only with an Array of Strings which has the following syntax:`[the remaining and appropriate new task or tasks on the language you have to use]`",
```
awtkns, line 32: @asim-shrestha tagging you here because i think you'll be interested with the prompts. Do you think this would significantly increase our costs?

awtkns (commented):
Looks great! Will take another look once the build passes!

public/locales/$LOCALE/$NAMESPACE.json/public/locales/ch/translation.json
Cs4K1Sr4C: Yes, I added this dir to the .gitignore; I will take a closer look on this issue why it has appeared (again)

src/utils/prompts.ts
Cs4K1Sr4C: @awtkns, @asim-shrestha
I can check the difference in tokens, but you can also do it. I tried to keep the length of the prompt by the way; my experiences shows that more declarative prompt will respond with a more efficient result.

awtkns:
Awesome, will review shortly!

PR #292, opened 2023-04-21
feat: add combobox for export options

Jshen123:
**Description:**
Export buttons are now grouped under one dropdown
image
**other**
- added Menu component


asim-shrestha:
Nice, 

we should ensure hover effects trigger on the entire group. Looks like hovering the drop down icon doesn't trigger the glow. Tailwind has [a group handler we can use for this.](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state)

We should also be able to click anywhere on the button. Looks like only the dropdown itself works currently

src/components/ChatWindow.tsx
```diff
@@ -167,6 +168,35 @@ const MacWindowHeader = (props: HeaderProps) => {
     void navigator.clipboard.writeText(text);
   };
 
+  const exportOptions = [
```
asim-shrestha, line 171: Should we extract an ExportComboBox element?

asim-shrestha:
> Nice,
> 
> we should ensure hover effects trigger on the entire group. Looks like hovering the drop down icon doesn't trigger the glow. Tailwind has [a group handler we can use for this.](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state)
> 
> We should also be able to click anywhere on the button. Looks like only the dropdown itself works currently

This might have to do with the fact that the other combobox is editable so some inconsistency here. TBH we probably don't even want it to be editable

Jshen123:
Hi @asim-shrestha, reimplemented the dropdown with the new menu component.
- button text should no longer be editable
- hover on chevron should trigger hover effect

awtkns:
Nice work here! Thoughts on keeping the save to db button outside the ddl? In doing that we could then rename it it to export.

src/components/ChatWindow.tsx
```diff
@@ -167,8 +168,34 @@ const MacWindowHeader = (props: HeaderProps) => {
...
+    <WindowButton
+      key="Agent"
```
Jshen123, line 173: @awtkns, I grouped save agent under the dropdown menu. 
-  it will be disabled unless  `onSave` is not undefined

awtkns (approved):
LTGM, my one suggestion would be to tighten up the padding on the save button, not sure it needs to quite so big.

Jshen123:
> Nice work here! Thoughts on keeping the save to db button outside the ddl? In doing that we could then rename it it to export.

I prefer it outside as well since it will be more obvious to the users.

It's mostly a responsive thing. not enough space for both when the screen is small. but I just have an idea now. taking a look

asim-shrestha:
We could also do a ping animation in the DDL for when export to DB is available: https://tailwindcss.com/docs/animation#ping

Jshen123:
@asim-shrestha @awtkns not my initial idea but i have separated out the Save agent button.

It's closer to what we have in production.

image

awtkns (approved):

awtkns:
@Jshen123 not sure why merging is blocked... i can force merge it for you if you want

PR #297, opened 2023-04-21
feat: Vector db integration

awtkns:


byronvoorbach:
We would love to add Weaviate to this PR as well @awtkns!

awtkns:
@byronvoorbach would you guys be able to help out with hosting 😉

hsm207:
> @byronvoorbach would you guys be able to help out with hosting 😉

@awtkns sure. we have [WCS](https://weaviate.io/pricing).

src/services/agent-service.ts
```diff
@@ -25,15 +26,25 @@ async function executeTaskAgent(
...
+  const ctx = await getContext({ id: "changeme", task });
+  const context = ctx.map((c) => c.document).join(", ");
+
   const completion = await new LLMChain({
     llm: createModel(modelSettings),
     prompt: executeTaskPrompt,
   }).call({
     goal,
     task,
+    context,
```
asim-shrestha, line 38: Do we need to pass in context for creating a task? Shouldn't this be for execute?

src/services/agent-service.ts
```diff
@@ -25,15 +26,25 @@ async function executeTaskAgent(
   goal: string,
   task: string
 ) {
+  const ctx = await getContext({ id: "changeme", task });
```
asim-shrestha, line 29: Should we use the run id as the id for the context?

Issue #261, opened 2023-04-19
🗄 Make left drawer collapsible even on large screens


asim-shrestha:
![image](https://user-images.githubusercontent.com/50181239/233157234-e1e97b98-6f7b-41b9-9206-83655f1af81e.png)

In small screens, this left drawer collapses. We should also do the same thing in larger screens, allowing for more screen space to read the Agent runs. Note the ChatWindow div currently has a max height, we should look into making this a larger value as well

2022H1030042G:
Hi, is the issue available? I would like to work on it.

asim-shrestha:
Yup! @2022H1030042G, gone ahead and assigned you

2022H1030042G:
Can you help me with this? I'm new to open source development.

2022H1030042G:
@rp-bot , While setting up the project, where do I have to create .env file as mentioned here:  https://github.com/reworkd/AgentGPT#:~:text=Create%20a%20.env%20file%20with%20the%20following%20content

2022H1030042G:
@rp-bot , Thanks for the information. Can I set it up on Windows?

2022H1030042G:
Okay, I created a .env file inside the Agent GPT folder and after running this command : https://github.com/reworkd/AgentGPT#:~:text=Ready,%2C%20now%20run%3A , I got this error: 
![image](https://user-images.githubusercontent.com/17452290/233639221-ebe8fb3c-d26a-44a5-87b5-a37f0efc878d.png)

2022H1030042G:
@rp-bot , @asim-shrestha , I am able to setup AgentGPT locally, now if can anyone help me with the changes that are required in a particular file.

2022H1030042G:
fix #307

PR #307, opened 2023-04-21
Drawer.tsx modified to make left drawer collapse on large screens

2022H1030042G:
Refer: #261

### DIFF
--- a/src/components/Drawer.tsx
+++ b/src/components/Drawer.tsx
@@ -29,6 +29,7 @@ const Drawer = ({
 }) => {
   const [showDrawer, setShowDrawer] = useState(false);
   const { session, signIn, signOut, status } = useAuth();
+  const [drawerWidth, setDrawerWidth] = useState(72);
   const router = useRouter();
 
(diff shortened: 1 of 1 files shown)


asim-shrestha:
Hey awesome start. Some notes though, even when the drawer is collapsed, the page remains fixed. We should be taking up the new space that was previously occupied by the drawer 

This also breaks mobile as the drawer button remains on the screen. Please look at the behaviour of the existing drawer on mobile for inspiration. (You can make a mobile view via chrome dev tools)

2022H1030042G:
> Hey awesome start. Some notes though, even when the drawer is collapsed, the page remains fixed. We should be taking up the new space that was previously occupied by the drawer
> 
> This also breaks mobile as the drawer button remains on the screen. Please look at the behaviour of the existing drawer on mobile for inspiration. (You can make a mobile view via chrome dev tools)

Actually, I'm new to open-source development, so if someone can help, that'd be great.

asim-shrestha:
Handled in #327, going to have to close this

PR #326, opened 2023-04-23
Fix Multi-Language Support

asim-shrestha:
- Translates FRONT END components only based on the user local. If the user is French for example, they will be transferred to /fr
- Adds language selector in the settings menu


.gitignore
```diff
@@ -41,4 +41,13 @@ yarn-error.log*
...
\ No newline at end of file
+.swc
+
+# extracted language files
+/public/locales/$LOCALES
+
+# package-lock.json
+#package-lock.json
+
+# .eslintcache
```
awtkns, line 52:
```suggestion
```

awtkns (approved):

public/locales/en/translation.json
```diff
@@ -0,0 +1,72 @@
...
+  "Please consider sponsoring the project on GitHub.": "Please consider sponsoring the project on GitHub.",
+  "SUPPORT_NOW": "Support now",
+  "Embarking on a new goal:": "Embarking on a new goal:",
+  "Added task:": "Added task:",
+  "Thinking...": "Thinking...",
+  "Executing:": "Executing:",
+  "Help": "Help",
+  "Settings": "Settings",
+  "Sign Out": "Sign Out",
+  "Sing In": "Sing In",
```
awtkns, line 17: This should be sign in.... unless we really do want to go the musical route 🎵

awtkns (commented):
image
We still may need to work on prompting in other languages

asim-shrestha:
> image
> 
> We still may need to work on prompting in other languages

Yeah haven't touched prompting just to keep this small

PR #332, opened 2023-04-24
Agent respond on the selected display language

Cs4K1Sr4C:
[mod] Agent respond on the selected display language
+ Selected language has the flag in the settings

### DIFF
--- a/src/components/SettingsDialog.tsx
+++ b/src/components/SettingsDialog.tsx
@@ -64,6 +64,7 @@ export const SettingsDialog: React.FC<{
   const handleReset = () => {
     customSettings.resetSettings();
     updateSettings("customApiKey", "");
+    close();
   };
 
   const disabled = !settings.customApiKey;
--- a/src/hooks/useSettings.ts
+++ b/src/hooks/useSettings.ts
@@ -8,7 +8,7 @@ import {
 
 const SETTINGS_KEY = "AGENTGPT_SETTINGS";
 const DEFAULT_SETTINGS: ModelSettings = {
-  customApiKey: undefined,
+  customApiKey: "",
   customModelName: GPT_35_TURBO,
   customTemperature: 0.9 as const,
   customMaxLoops: DEFAULT_MAX_LOOPS_CUSTOM_API_KEY,


src/utils/prompts.ts
```diff
@@ -17,20 +17,58 @@ export const createModel = (settings: ModelSettings) => {
...
+  inputVariables: ["goal", "task", "language"],
+});
+
+export const createTasksPrompt = new PromptTemplate({
+  template:
+    "You're AgentGPT, a super-intelligent AI tasked with creating and completing objectives. You must follow the provided syntax: 1. Include {language} in your response language. 2. Your main objective is {goal}. 3. Your incomplete tasks are: {tasks}. 4. Your last completed task was: {lastTask}. 5. The result of your last task was: {result}. Create new tasks as needed to achieve your objective based on your previous and incomplete tasks. Respond only with an array of strings that can be processed by JSON.parse().",
+  inputVariables: ["goal", "tasks", "lastTask", "result", "language"],
+});
```
asim-shrestha, line 37: This style of prompting might be confusing

image

For example, it starts of by specifying "English" which isn't really relevant for users

src/utils/prompts.ts
```diff
@@ -17,20 +17,58 @@ export const createModel = (settings: ModelSettings) => {
...
+
+export const startGoalPrompt = new PromptTemplate({
+  template:
+    "As AgentGPT, you're an AI designed to create tasks and complete objectives using predefined response syntax. To complete a given objective, follow these rules: 1. Set your response language to {language}. 2. Analyze the objective: {goal}. Respond with an array of strings enclosed with braclets: [1...3 tasks in {language}]",
+  inputVariables: ["goal", "language"],
+});
+
+export const executeTaskPrompt = new PromptTemplate({
+  template:
+    "You're AgentGPT, a highly intelligent AI that creates and completes tasks. Your responses follow a predefined syntax, and you must be logical and attentive to complete your objectives. Remember to: 1) Specify the language of your response with {language}. 2) State your current main objective as {goal}. 3) Report your current task in progress as {task}. Once you've analyzed and executed your current task, provide the result or solution in a single string using the following syntax: '{{the result or solution on {language}}}'",
```
asim-shrestha, line 29: Lets use the new response format specifications but keep the prompts similar otherwise, we need a better suite of tests or understanding of how these prompts affect output before we can change them too much

src/components/LanguageCombobox.tsx
```diff
@@ -15,7 +15,9 @@ const LanguageCombobox = () => {
 
   const handleInputChange = (languageName: string) => {
     const selectedLanguage = findLanguage(languageName);
+    i18n.changeLanguage(selectedLanguage.code);
     setActualLanguage(selectedLanguage);
+    console.log(i18n.language);
```
asim-shrestha, line 20: remove log

src/utils/helpers.ts
```diff
@@ -37,13 +37,32 @@ export const extractArray = (inputStr: string): string[] => {
...
+            // Enclose the input string with braces and try to parse again
+            const enclosedStr = `{[${inputStr}]}`;
+            return JSON.parse(enclosedStr) as string[];
+          } catch (bracesError) {
+            if (bracesError instanceof Error) {
+              console.error("Error parsing the enclosed array with braces:", bracesError);
+            }
+          }
+        }
+      }
```
asim-shrestha, line 57: What's the use case here? Don't know of scenarios where it returns a CSV of strings without the array exterior

src/pages/index.tsx
```diff
@@ -226,7 +245,11 @@ const Home: NextPage = () => {
             <Expand delay={1.4} className="flex gap-2">
               <Button
                 disabled={disableDeployAgent}
-                onClick={handleNewGoal}
+                onClick={async () => {
+                  await setAgentLanguage(findLanguage(i18n.language)["name"]);
+                  console.log(agentLanguage);
+                  handleNewGoal();
+                }}
```
asim-shrestha, line 252: handleNewGoal recreates the agent. We don't need to set the language manually

src/pages/index.tsx
```diff
@@ -35,6 +39,15 @@ const Home: NextPage = () => {
   const [showSettingsDialog, setShowSettingsDialog] = React.useState(false);
   const [hasSaved, setHasSaved] = React.useState(false);
   const agentUtils = useAgent();
+  const router = useRouter();
+  const { locale } = router;
```
asim-shrestha, line 43: The router locale value will change as the language changes, I don't think we need both i18n hook and the agent language as a state variable. Just a findLanguage function from the locale that we pass in when we create the agent itself.

asim-shrestha (changes requested):

PR #340, opened 2023-04-24
Zjini patch 1

Zjini:


asim-shrestha:
Hey what's the use case here? We shouldn't be removing sponsors

PR #342, opened 2023-04-24
feat: guest mode

Dogtiti:
This feature is designed for scenarios that do not require back-end service login. Some users may wish to deploy their own web pages while also providing services to friends and associates without exposing their Open AI key. In such cases, a "Guest Key" can be set up for friends to use. Once this key has been set, the visitor mode is automatically enabled. Your friend must trigger the input of their own Open AI key or access code in the settings panel of their page before they can use the service.

I have made initial progress on this feature, but there may still be some details that require optimization.

The key should be in this format: `abc,qwe,123`, where each comma-separated value can be used.

### DIFF
--- a/src/components/AutonomousAgent.ts
+++ b/src/components/AutonomousAgent.ts
@@ -27,6 +27,7 @@ class AutonomousAgent {
   numLoops = 0;
   session?: Session;
   _id: string;
+  isValidGuest = false;
 
   constructor(
     name: string,
@@ -34,6 +35,7 @@ class AutonomousAgent {
     renderMessage: (message: Message) => void,
     shutdown: () => void,
     modelSettings: ModelSettings,
+    isValidGuest: boolean,
     session?: Session
   ) {
     this.name = name;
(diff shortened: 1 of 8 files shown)


asim-shrestha:
Can you explain the use case for a guest? What web pages / services would people be deploying?

Dogtiti:
> Can you explain the use case for a guest? What web pages / services would people be deploying?

The use case for a guest is when someone has deployed an app, and they want their friends to be able to use it, but not anyone else. Without providing their friends with a key, they can provide a guest key for their friends to use, so they can use it without exposing their key (assuming their friends don't have a key and don't want to apply for one, but just want to use it).

awtkns:
@Dogtiti I am not sure this is a direction we want to be going in. If we were to have some sort of anonymous authentication then it would be important for us to have it integrated with NEXT_AUTH.

Dogtiti:
@awtkns You may be right, but there are many people who may not be using next-auth as it may be too heavy for personal use, or they may prefer a one-click deployment on Vercel without any additional configurations such as auth. However, I would still appreciate your suggestions and any good ideas on how to handle this scenario.

asim-shrestha:
Why not just make a guest key in OpenAI and have them use it directly, or using it as the OpenAI env key directly, or having your friend(s) deploy themselves?

Dogtiti:
Because if you provide an OpenAI key, it can be used anywhere, but a guest key can only be used for the application it was deployed for. However, if this doesn't fit with your direction, I'll turn it off. If someday you have a similar need, you can reopen it and I can offer some assistance.

PR #359, opened 2023-04-26
Docker-Compose and Interactive configuration script

JPDucky:
Created a setup script and a docker-compose route to make it easier for beginners. Currently only docker-compose is working, no errors. I will need to do more work on docker and local build tomorrow. Documentation to follow. 

By calling the init.sh file, it installs the required packages for the interactive installation script, which then asks the user if what parameters they want to set, taken from setup.sh. The defaults are set with those suggested in setup.sh, and by not entering anything and hitting 'enter' they will be entered. It then asks the user if they want to deploy via docker-compose, docker, or node.

If the user chooses 'docker-compose' it will automatically build a docker compose file and launch it. This is currently the only option that works properly. Actually, now that I think of it it builds a docker compose file regardless. I will fix that tomorrow and will get the other two working as well. 

I also need to set up the script to not launch the configurator if the environment files/docker-compose files exist.

I have tested the docker-compose route thoroughly and it works without issues. It's late and I'm exhausted so I will continue work on this in the morning. Cheers.

image


asim-shrestha:
Awesome @JPDucky, a bit cloudy due to the merge, would you mind running another merge to avoid these fake diffs?

asim-shrestha:
Closing in favor of #406

PR #371, opened 2023-04-26
Feat/update task status

Jshen123:


Jshen123:
This change requires DB update

PR #383, opened 2023-04-27
Alexk1919 patch 1

alexk1919:
Remove extra div in HelpDialog.tsx


asim-shrestha:
Heyo, so this leads to the following error

```
Type error: JSX element 'div' has no corresponding closing tag.
Expand 8 Lines
18:11:03.457
```

It doesn't seem like we have an extra div here, but if you can also delete the corresponding unnecessary opening tag then we're happy to merge this

alexk1919:
Hi Asim, I'll close this PR. On my end, this extra div was causing a deploy issue on Vercel. It seems on your end, it is working fine.

PR #385, opened 2023-04-28
feat: implement pause functionality (WIP)

Jshen123:
Description
- implement global store for agent ✅ 
- implement pause functionality 🚧


src/components/ChatWindow.tsx
Jshen123, line 1: `isAgentStopped` was never needed in `ChatWindow`

src/components/TaskWindow.tsx
```diff
@@ -18,21 +19,16 @@ export const TaskWindow = ({ isAgentStopped }: { isAgentStopped: boolean }) => {
...
+            <Task key={i} task={task} />
           ))}
         </div>
       </div>
     </Expand>
   );
 };
 
-const Task = ({
-  task,
```
Jshen123, line 30: use zustand global store and avoid prop drilling

PR #393, opened 2023-04-28
feat: add endpoint url setting

Dogtiti:


asim-shrestha:
Hey @Dogtiti I'm not sure we can support this for security reasons. It might be possible for a bad actor to supply a URL that would just store our API key for an example. CC @awtkns

asim-shrestha:
Closing because of the above issue, happy to do a PR where it takes in an ENV endpoint for local dev though

PR #400, opened 2023-04-29
feat: implement agent global state

Jshen123:


src/components/stores/agentStore.ts
```diff
@@ -0,0 +1,39 @@
...
+const initialAgentState = {
+  agent: null,
+  isAgentStopped: true,
+};
+
+interface AgentSlice {
+  agent: AutonomousAgent | null;
+  isAgentStopped: boolean;
+  setIsAgentStopped: () => void;
+  setAgent: (newAgent: AutonomousAgent | null) => void;
```
asim-shrestha, line 15: an interface of 

```
removeAgent: () => void;
setAgent: (newAgent: AutonomousAgent) => void;
```

Might make more sense. Remove agent would agent.stop(), setAgent(null) and also update isAgentStopped

src/pages/index.tsx
```diff
@@ -19,25 +19,34 @@ import { useAuth } from "../hooks/useAuth";
...
+  const setIsAgentStopped = useAgentStore.use.setIsAgentStopped();
+  const agent = useAgentStore.use.agent();
+
   const { session, status } = useAuth();
   const [name, setName] = React.useState<string>("");
   const [goalInput, setGoalInput] = React.useState<string>("");
-  const [agent, setAgent] = React.useState<AutonomousAgent | null>(null);
   const settingsModel = useSettings();
-  const [shouldAgentStop, setShouldAgentStop] = React.useState(false);
+  // const [shouldAgentStop, setShouldAgentStop] = React.useState(false);
```
asim-shrestha, line 48: delete?

src/pages/index.tsx
```diff
@@ -115,7 +114,6 @@ const Home: NextPage = () => {
 
   const handleStopAgent = () => {
     agent?.stopAgent();
-    setIsAgentStopped();
```
Jshen123: @asim-shrestha  removed `setIsAgentStopped()` since there is already a useEffect hook that update `IsAgentStopped`

https://github.com/reworkd/AgentGPT/blob/6c5880228620cbbd31306be5725e0be19d42821f/src/pages/index.tsx#L72-L74

src/types/agentTypes.ts
```diff
@@ -65,6 +66,20 @@ export const messageSchema = z.union([taskSchema, nonTaskScehma]);
...
+];
+
+const AgentStatusSchema = z.union([
+  z.literal(AGENT_STATUS_RUNNING),
+  z.literal(AGENT_STATUS_STOPPED),
+  z.literal(""),
+]);
+
+export type AgentStatus = z.infer<typeof AgentStatusSchema>;
+
```
awtkns, line 82: This seems a hard to work with. Any reason we cant do the below?
```suggestion
export type AgentStatus = "running" | "stopped"
```

src/pages/index.tsx
```diff
@@ -248,11 +248,11 @@ const Home: NextPage = () => {
                 )}
               </Button>
               <Button
-                disabled={agent == null}
+                disabled={agent === null}
                 onClick={handleStopAgent}
                 enabledClassName={"bg-red-600 hover:bg-red-400"}
               >
-                {shouldAgentStop ? (
+                {!isAgentStopped && agent === null ? (
```
awtkns, line 255: Not sure if this makes semantic sense. The above is saying:
"If the agent is not stopped and the agent exists then show stopping" 
^ not sure that is right semantically

awtkns (commented):
looking good! I will wait for @asim-shrestha though

src/types/agentTypes.ts
Jshen123: mostly because the strings "running" and "stopped" is also used in many places in the code

src/pages/index.tsx
Jshen123: agreed, I could be wrong but this might be outdated code and imo we should just remove this completely. 

my understanding is that this is handling a state where agent has been shutdown (i.e. set to `null`) but the `agent?.isRunning` has not been set to `false`

In today's code, `agent?.isRunning` and `setAgent(null)` happen simultaneous. I don't think this state would happen
https://github.com/reworkd/AgentGPT/blob/4e83cd0b160396c5fec1cd1fff4f6a0fb94533a6/src/components/AutonomousAgent.ts#L255-L260

CC: @asim-shrestha

src/types/agentTypes.ts
Jshen123: but on 2nd thought, I can make this more concise.

src/pages/index.tsx
asim-shrestha: Yeah lets refactor this after

PR #413, opened 2023-04-30
feat: implement pause functionality

Jshen123:
**Description:**
New Pause Mode can now be enabled in settings

image


awtkns:
Instead of having this hidden deep in the settings what if we had either a button either on the terminal or in the task itself? Seems strange that wee need an api key to enable this. 

cc @nif33 not sure if you had thoughts about this

Jshen123:
> Instead of having this hidden deep in the settings

Good call @awtkns! Moved `Mode` up into basic settings

> Seems strange that wee need an api key to enable this

It's dependent on the existence of an agent instance


Couple of thoughts regarding moving `mode` setting out :) My initial instinct was also to display it on the same page as the terminal (I even made a switch component for this) but I ultimately decided against it:
1. There isn't really an ideal place on the terminal page for this component on a mobile screen.
2. The frequency of interaction for this component is likely very low (I imagine the median user would probably touch this once or twice).  When steerability matures and becomes stable in the future, pause/interact will probably become the default mode
3. Future-proofing: if we were to move this to the terminal page, this will probably end up becoming some sort of binary toggle. Although I don't currently foresee a third/fourth mode, I don't want the design to restrict that possibility.

asim-shrestha:
Nice @Jshen123, works really well 🙏 

> My initial instinct was also to display it on the same page as the terminal

I think we should start out with this (otherwise most people will miss it). Would also allow for quick user experimentation as they use the tool. I'm also thinking of adding a toggle for web search as that is still experimental.

> There isn't really an ideal place on the terminal page for this component on a mobile screen.

We should be able to make some space for a switch section. It might feel slightly off at first but we can iterate on it.

> This will probably end up becoming some sort of binary toggle. Although I don't currently foresee a third/fourth mode, I don't want the design to restrict that possibility.

The toggle can just flip between the two modes while the component itself could allow for multiple modes. But TBH for something like this we probably shouldn't care to future proof and should deal with those cases as needed 

> Seems strange that wee need an api key to enable this
...
It's dependent on the existence of an agent instance

We should only require an API key for settings that will directly affect our billing (# loops, model choice, # tokens) or are very very closely related to OpenAI apis (temperature). In this case, this is outside of both of those things and just dictates how we run agents

> Good call awtkns@! Moved Mode up into basic settings

Speaking of settings, we should Hide advanced settings (or make it unopenable) without providing an API key. The model choice should also be placed within advanced settings. We should then allow for saving when no API key is provided. (All of this should be handled in a separate PR though, I can make some issues)

asim-shrestha:
image

image

Example

awtkns:
Pause button at the bottom right of the terminal also feels pretty natural to me imho

src/pages/index.tsx
```diff
@@ -234,16 +278,7 @@ const Home: NextPage = () => {
...
-                {agent == null ? (
-                  t("Deploy Agent")
-                ) : (
-                  <>
-                    <VscLoading className="animate-spin" size={20} />
-                    <span className="ml-2">{t("Running")}</span>
-                  </>
-                )}
-              </Button>
+              {firstButton}
```
asim-shrestha, line 281: Maybe worth just extracting buttons to a component

Could also rename to deployment button

src/pages/index.tsx
```diff
@@ -99,21 +112,34 @@ const Home: NextPage = () => {
...
 
+  const handleContinue = () => {
+    if (!agent) {
+      return;
+    }
+
+    agent.updatePlayBackControl(AGENT_PLAY);
+    updateIsAgentPaused(agent.playbackControl);
+    agent.updateIsRunning(true);
+    agent.run().then(console.log).catch(console.error);
```
asim-shrestha, line 123: We shouldn't have to manage both the global state and the agent state individually. We should just have zustand functions that will update the global state and manually deal with the agent under the hood

src/pages/index.tsx
```diff
@@ -81,6 +84,14 @@ const Home: NextPage = () => {
     addMessage(message);
   };
 
+  const handlePause = (opts: {
+    agentPlaybackControl?: AgentPlaybackControl;
+  }) => {
+    if (opts.agentPlaybackControl !== undefined) {
+      updateIsAgentPaused(opts.agentPlaybackControl);
```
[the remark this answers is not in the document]
asim-shrestha: And should abstract away all intricacies of pause / play / etc to zustand

src/components/AutonomousAgent.ts
```diff
@@ -33,60 +42,79 @@ class AutonomousAgent {
...
-      console.log(e);
-      this.sendErrorMessage(getMessageFromError(e));
-      this.shutdown();
-      return;
     }
 
     await this.loop();
+    if (this.mode === PAUSE_MODE && !this.isRunning) {
+      this.handlePause({ agentPlaybackControl: this.playbackControl });
+    }
```
asim-shrestha, line 109: What is this for?

src/components/AutonomousAgent.ts
```diff
@@ -143,21 +171,32 @@ class AutonomousAgent {
...
       );
-      currentTask.status = TASK_STATUS_FINAL;
-      this.sendMessage(currentTask);
+      this.sendMessage({ ...currentTask, status: TASK_STATUS_FINAL });
     }
-
     await this.loop();
   }
 
+  private conditionalPause() {
```
asim-shrestha, line 186: `handlePause` might be better

asim-shrestha (approved):

src/components/AutonomousAgent.ts
Jshen123: this to inform the components whether the Agent is Paused or Running (for example, so it knows whether to display Continue button or not)

src/styles/globals.css
```diff
@@ -96,3 +96,14 @@ div::-webkit-scrollbar-thumb {
...
+/*  disappearing animation */
+.animation-hide {
+  animation: hide 1s cubic-bezier(0.4, 0, 0.2, 1) 0s 1 forwards;
+}
+
+@keyframes hide {
+  to {
+    opacity: 0;
+  }
+}
```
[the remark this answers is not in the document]
Jshen123: will take a look

src/pages/index.tsx
Jshen123: Like encapsulating the above logic under zustand action? I can do that.

src/pages/index.tsx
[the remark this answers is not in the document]
Jshen123: yeah, this is sort of a hack. I had some issue detecting instance variables. Will take a look as part of enhancement

PR #419, opened 2023-04-30
feat: Remove thinking tasks after completion

awtkns:
Ensures that we only ever show one thinking task at once. Should also mean we are not saving them in the db now. Will do a serverside followup later :)

(i have millions of thinking tasks i want to clean from db)

### DIFF
--- a/src/components/stores/messageStore.ts
+++ b/src/components/stores/messageStore.ts
@@ -26,6 +26,16 @@ interface MessageSlice {
   addMessage: (newMessage: Message) => void;
 }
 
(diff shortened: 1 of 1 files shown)


awtkns:
cc @Jshen123 as this is my first time with zustand

awtkns:
Seems like it would be good at add a thinking message after task creation to. eg; 
Screen Shot 2023-04-30 at 10 17 02 PM

This can come later though. @asim-shrestha please feel free to merge this if you think it is good.

asim-shrestha:
There should be thinking steps in between, I suspect this is removing them before they display

PR #437, opened 2023-05-03
📃 Add examples to the front page and fix prompts

asim-shrestha:
image


awtkns (approved):

awtkns:
@asim-shrestha failing to build it seems other then that it look good. For the example I think we should be really loading these from the database in the future as they will be fairly static. This could help have some OPEN ai costs.

PR #438, opened 2023-05-03
feat: drag and drop task window

### LINKED ISSUE #113: Feature Request: Have an individual task execute button
Not sure if this is in the roadmap, but it would be great if we could execute specific tasks from the lists one at a time manually as well. This would allow us to filter tasks that we dont find relevant and get some insights.

Jacky-He:
closes #113 
- Add the drag and drop functionality to task window so that the order of tasks can be changed
Screenshot 2023-05-06 at 10 43 11 PM


prisma/schema.prismae
```diff
@@ -0,0 +1,99 @@
+generator client {
```
awtkns, line 1: This file should not be committed

prisma/schema.prisma
```diff
@@ -3,7 +3,7 @@ generator client {
 }
 
 datasource db {
-  provider     = "mysql"
+  provider     = "sqlite"
```
awtkns, line 6: The changes to this file should not be committed :)

package.json
```diff
@@ -16,6 +16,7 @@
   },
   "dependencies": {
     "@headlessui/react": "^1.7.14",
+    "@hello-pangea/dnd": "^16.2.0",
```
awtkns, line 19: Instead of adding another animation library, what are your thoughts on using framer-motion instead as this is already bundled as dependancy. Eg: https://codesandbox.io/s/framer-motion-drag-to-reorder-pkm1k?file=/src/Example.tsx

awtkns (commented):
Looking good so far, just left a few comments. Amazing work 🚀 

Also seems there is a build error and some merge conflicts.

package.json
Jacky-He: Hi @awtkns thanks for the comments! We think that @hello-pangea/dnd is easier for us to work with for the moment. We do think that "framer-motion" is more flexible and seems to be a cool library to use, so it could be good for a future enhancement task. Does that sound good to you?

Jshen123 (changes requested):
I think this is a cool idea :) but have some feedbacks and concerns about this PR:

1. From an UI perspective, the order of **Current Tasks**  currently implies a chronological progression. 
a. If we allow drag and drop, we need to decide what the order of  **Current Tasks** is trying to communicate.
b. We also need to design rules and requirements on what type of task can be dragged above other tasks (e.g. we probably shouldn't allow started task to be dragged above a completed task)
2. Right now, it looks like only `tasks` state is being arranged. I think the `messages` state also need to be correctly rearranged to be in sync with reordered tasks.
a. This brings up another question. Should we have this type of mutation for messages? (similar concern as 1, Messages currently serves a chronological log of the agent's activity)
3. I think `reorderTasks` will create issues if run on **automatic mode**. Right now it looks like it's setting the `tasks` state with an in-memory state `updatedTasks`. However, because the `AutonomousAgent` is running in parallel, any task generated by `AutonomousAgent` during the reorder operation will be overwritten.

daga-fam:
Does this PR include running the tasks manually as well?

PR #460, opened 2023-05-06
Add tasks

asim-shrestha:
- Fix wonky scroll bar in task window
- Add an input and button in task window to add a custom task
- Once deployed, keep the task window displaying

image

### DIFF
--- a/src/components/Input.tsx
+++ b/src/components/Input.tsx
@@ -6,7 +6,7 @@ import { isArrayOfType } from "../utils/helpers";
 import type { toolTipProperties } from "./types";
 
 interface InputProps {
-  className?: string;
+  small?: boolean; // Will lower padding and font size. Currently only works for the default input
   left?: React.ReactNode;
   value: string | number | undefined;
   onChange: (
@@ -31,7 +31,7 @@ interface InputProps {
 
 const Input = (props: InputProps) => {
   const {
-    className,
+    small,
     placeholder,
     left,
     value,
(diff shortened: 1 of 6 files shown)


src/components/Input.tsx
```diff
@@ -104,7 +106,8 @@ const Input = (props: InputProps) => {
           "border:black delay-50 w-full rounded-xl bg-[#3a3a3a] py-1 text-sm tracking-wider outline-0 transition-all placeholder:text-white/20 hover:border-[#1E88E5]/40 focus:border-[#1E88E5] sm:py-3 md:text-lg",
           !isTypeRange() && "border-[2px] border-white/10 px-2",
           disabled && " cursor-not-allowed hover:border-white/10",
-          left && "md:rounded-l-none"
+          left && "md:rounded-l-none",
+          className
```
awtkns, line 110: nice!

src/components/TaskWindow.tsx
```diff
@@ -1,28 +1,63 @@
...
   const tasks = useMessageStore.use.tasks();
+  const addMessage = useMessageStore.use.addMessage();
   const [t] = useTranslation();
+
+  const handleAddTask = () => {
+    addMessage({
+      taskId: v1().toString(),
+      value: customTask,
+      status: TASK_STATUS_STARTED,
+      type: MESSAGE_TYPE_TASK,
```
awtkns, line 30: More of a convo for a later when we develop a style guide, Curious if you're partial to:
```ts
      status: TASK_STATUS_STARTED,
      type: MESSAGE_TYPE_TASK,
      
      //or
      status: "status",
      type: "task",
```
With typescript i am more a fan of that later but can see the benefits of both

awtkns (approved):

src/components/TaskWindow.tsx
asim-shrestha: latter is probably easier, thoughts on Enums?

PR #469, opened 2023-05-07
fix: move mode toggle out of settings

Jshen123:
**Description:**
No longer inside settings
image


src/components/ChatWindow.tsx
```diff
@@ -167,22 +174,42 @@ const ChatWindow = ({
...
-                value={isWebSearchEnabled}
-                onChange={handleChangeWebSearch}
-              />
-            </div>
-          </div>
-        </>
+        <div className="flex flex-col items-center justify-center md:flex-row">
+          <SwitchContainer label="Web Search">
+            <Switch
+              disabled={agent !== null}
```
asim-shrestha, line 180: Shouldn't be disabled

Issue #450, opened 2023-05-05
TypeError: Cannot read properties of undefined (reading 'Unicode')


eeezae:
<img width="946" alt="image" src="https://user-images.githubusercontent.com/34007837/236365686-e28f9ef7-fc93-4c10-abc7-a28717346cce.png">

gsherofly007:
me too

gsherofly007:
I wait for the developer to solve the problem

awtkns:
cc @Cs4K1Sr4C this looks related to the pdf changes you made

awtkns:
Awesome @Cs4K1Sr4C! I've assigned it to you for now, let me know if you need any help. 😀

awtkns:
fixed in #465

asim-shrestha:
Closing as we haven't heard of any other issues

gsherofly007:
您好，邮件收到

PR #470, opened 2023-05-07
[fix] PDF export

Cs4K1Sr4C:
Fix for #450 

PDF export supports the following special languages with line wrapping:
- Chinese
- Japanese
- Korean
- Russian

image

image

image

image


asim-shrestha:
These fonts are extremely large, I think it might cause problems on the server. CC @awtkns

src/components/pdf/PDFButton.tsx
```diff
@@ -1,8 +1,17 @@
...
+import ReactPDF, {
+  Document,
+  Font,
+  Page,
+  pdf,
+  StyleSheet,
+  Text,
+} from "@react-pdf/renderer";
+import View = ReactPDF.View;
+import jsPDF from "jspdf";
```
asim-shrestha, line 12: Is this being used still? Can we remove it + remove it from package.json?

public/locales/hr/common.json
```diff
@@ -1,20 +1,20 @@
 {
 	"ADDING_TASK": "Dodavanje zadatka",
 	"AGENTGPT_DOCUMENTATION": "Dokumentacija AgentGPT-a",
-	"CLOSE": "Zatvori",
+	"CLOSE": "Zatvoriti",
 	"CONTINUE": "Nastaviti",
-	"COPIED_TO_CLIPBOARD": "Kopirano u međuspremnik! \n🚀",
-	"COPY": "Kopiraj",
+	"COPIED_TO_CLIPBOARD": "Kopirano u međuspremnik! 🚀",
```
asim-shrestha, line 6: We should avoid adding emojis to the translations and just leave them in the string

PR #482, opened 2023-05-08
Add Traditional Chinese Translation + edit Simplified Chinese missing translation 

enjuichang:
I've added and edited some translations for these two languages. Let me now if I need to change anything!


asim-shrestha:
Appreciate it @enjuichang! Tidy-ed it up a bit, biggest thing was we shouldn't commit the schema updates

Also, it seemed like the "-" in `zh-tw` was causing issues so i replaced it with just `zhtw`

PR #485, opened 2023-05-09
📄 Add buttons to display TaskWindow on mobile

asim-shrestha:
## Notes
The task window currently is only visible on screen size `xl` and above. This means mobile users will be void of a decent bit of app functionality. This PR does the following:

- Make TaskWindow always visible instead of only appearing when tasks are present
- Add a button group to the top of the page when screen size is lower than `xl`
- The button group controls which window is visible. This defaults to the chat window

## Mobile
image
image

## Web (Smaller than XL)
image
image

### DIFF
--- a/src/components/ChatWindow.tsx
+++ b/src/components/ChatWindow.tsx
@@ -40,6 +40,7 @@ interface ChatWindowProps extends HeaderProps {
   displaySettings?: boolean; // Controls if settings are displayed at the bottom of the ChatWindow
   openSorryDialog?: () => void;
   setAgentRun?: (name: string, goal: string) => void;
+  visibleOnMobile?: boolean;
 }
 
 const messageListId = "chat-window-message-list";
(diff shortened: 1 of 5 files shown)


src/components/ChatWindow.tsx
```diff
@@ -11,17 +11,17 @@ import FadeIn from "./motions/FadeIn";
...
-  TASK_STATUS_STARTED,
-  TASK_STATUS_EXECUTING,
+  MESSAGE_TYPE_THINKING,
+  PAUSE_MODE,
   TASK_STATUS_COMPLETED,
+  TASK_STATUS_EXECUTING,
   TASK_STATUS_FINAL,
-  AUTOMATIC_MODE,
-  PAUSE_MODE,
+  TASK_STATUS_STARTED,
```
asim-shrestha, line 24: From linter

PR #492, opened 2023-05-10
🐋 Put MySQL DB in docker

asim-shrestha:
## Notes
- Add MySQL container into docker-compose
- Make schema use MySQL again
- Make `platform` only use SSL certs if it is in production
- Add a script so that the front end waits for mysql availability before running prisma migrations
- Make next Dockerfile run in dev mode


platform/reworkd_platform/db/utils.py
```diff
@@ -12,6 +12,12 @@ def create_engine() -> AsyncEngine:
 
     :return: SQLAlchemy engine instance.
     """
+    if settings.environment == "dev":
```
awtkns, line 15: to keep it consistent the environment should be `developement` (what vercel uses)

Issue #508, opened 2023-05-11
Invalid environment variables: NEXT_PUBLIC_BACKEND_URL: Required


SnoopyDevelops:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/reworkd/AgentGPT/labels/bug) didn't find any similar reports.

### Expected Behavior

The setup script should be run without problem.

### Current behaviour

![image](https://github.com/reworkd/AgentGPT/assets/54778084/616f2e05-9004-4e3d-bc0f-231601006e3a)

### Steps to reproduce

1. Create a new Codespace.
2. run `bash ./setup.sh --local` in terminal.

### Which Operating Systems are you using?

- [ ] Android
- [ ] iPhone/iPad
- [X] Linux
- [ ] macOS
- [ ] Windows

### Acknowledgements

- [X] My issue title is concise, descriptive, and in title casing.
- [X] I have searched the existing issues to make sure this bug has not been reported yet.
- [X] I am using the latest version of AgentGPT.
- [X] I have provided enough information for the maintainers to reproduce and diagnose the issue.

linghongli:
Same question

tom-dell:
Could you try setting a environment variable to the one found in .env.example? 

`export NEXT_PUBLIC_BACKEND_URL=http://localhost:3000`

Or maybe appending it to /workspaces/agentgpt/next/.env

SnoopyDevelops:
Added NEXT_PUBLIC_BACKEND_URL in the environment variable section of setup.sh, then it fixed the error

asim-shrestha:
Thanks @SnoopyDevelops !

PR #509, opened 2023-05-11
Fix Invalid environment variables(NEXT_PUBLIC_BACKEND_URL: Required)

SnoopyDevelops:
https://github.com/reworkd/AgentGPT/issues/508

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -22,6 +22,7 @@ NEXTAUTH_SECRET=$(openssl rand -base64 32)
 ENV="NODE_ENV=development\n\
 NEXTAUTH_SECRET=$NEXTAUTH_SECRET\n\
 NEXTAUTH_URL=http://localhost:3000\n\
+NEXT_PUBLIC_BACKEND_URL=http://localhost:3000\n\
 OPENAI_API_KEY=$OPENAI_API_KEY\n\
 DATABASE_URL=file:../db/db.sqlite\n"
 


setup.sh
```diff
@@ -22,6 +22,7 @@ NEXTAUTH_SECRET=$(openssl rand -base64 32)
 ENV="NODE_ENV=development\n\
 NEXTAUTH_SECRET=$NEXTAUTH_SECRET\n\
 NEXTAUTH_URL=http://localhost:3000\n\
+NEXT_PUBLIC_BACKEND_URL=http://localhost:3000\n\
 OPENAI_API_KEY=$OPENAI_API_KEY\n\
```
asim-shrestha, line 26: We should remove the trailing slash. Good otherwise

SnoopyDevelops: It's not added as the last environment variable.
Still the trailing slash should be removed?

SnoopyDevelops: @asim-shrestha 
It's inserted before the OPENAI_API_KEY and all other environmental variables have the trailing slash except the last one.
Still the trailing slash should be removed when it's not the last environment variable?

awtkns: @SnoopyDevelops @asim-shrestha I think the trailing slash is needed here. Looks good to me 🚀

awtkns (approved):

setup.sh
asim-shrestha: Ooops apologies, didn't realize it was needed for the new line

PR #532, opened 2023-05-14
🔥 Auto-translate via Google (free) or OpenAI (needs API key)

Cs4K1Sr4C:
- Collects and translates the `translate` functions
- Default source translation JSON files from `public/locales/en` directory
- Must to have `OPENAI_API_KEY` in ENV if the `TRANSLATOR_SERVICE` is set to `openai`
- Alternative translation using Google translate (doesn't need any API Key)
- If there is no namespace provided in the `translate` function then uses the `common` as default
- Possible translation methods: chat, text
- Possible translation services: google, openai
- Create a separated directory in the `public/locales/` named "translation" to avoid overwrite or delete already existing translations
- Has 3 modes: Collect and Translate, Only Collect, Only Translate
- Accept parameters and automatically parse them if they are provided


awtkns:
This is awesome @Cs4K1Sr4C!

next/translate.cjs
```diff
@@ -81,9 +81,38 @@ const collectAndTranslate = async (
...
+            console.log(translationKey + " added to " + targetTranslationFile);
+          }
+        }
+      }
+    }
+  });
+};
 
-const onlyTranslate = async () => {};
+const onlyTranslate = async (dirPath, sourceLanguage, targetLanguage) => {};
```
Cs4K1Sr4C, line 115: Next PR...

Cs4K1Sr4C (commented):
It still seems good... 👍

Cs4K1Sr4C:
So if we have a new static text to translate just easily use the `translate("{key}", "{default text in English}", "{?namespace?}")` syntax. After if you run the translator script (`npm translate`) it will find and process the mentioned formula in every .ts and .tsx files; after then it will extract then the  found `key/text/namespace` combinations into an array then translates the text to the selected target language; During the search and extraction process it will change the source code of the file where from the translatable part has been extracted to the single `translate("{key}", "{?namespace?}")` syntax.

asim-shrestha:
Doesn't really belong in the main repo. Closing as a result, you can find this project here: https://github.com/Cs4K1Sr4C/ATi18n

PR #542, opened 2023-05-15
🚀 Lighter, more transparent theme with improved shadows on the Drawer and on the ChatWindow

Cs4K1Sr4C:
If somebody working on something like this with a toggle button  then excuse me for creating this PR. I just made it because sometimes my eyes need some changes in contrasts. Maybe I'm not the only one...

**Lighter**

image

image

**Current**

image

image


Cs4K1Sr4C:
Close due to upstream more updated.
Will turn back with an other and updated concept.

PR #543, opened 2023-05-15
✨ Support for custom base URL

asim-shrestha:


Swiffers:
I'm setting export OPENAI_API_BASE="http://localhost:8000", but it's still using OpenAI Servers. Am I doing something wrong?

PR #561, opened 2023-05-18
Vercel setup

alexmaster1998:


awtkns (changes requested):
Hey @alexmaster1998 looks like the prisma file was added accidentally. Other than that this looks good.

PR #650, opened 2023-06-01
🧠 Basics of memory with Weaviate

asim-shrestha:


platform/reworkd_platform/web/api/agent/memory/memory.py
```diff
@@ -0,0 +1,71 @@
...
+
+        # Return
+        return [
+            (text.page_content, score)
+            for [text, score] in results
+            if score > similarity_threshold
+        ]
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.client.__del__()
```
asim-shrestha, line 71: Put this after __enter__ ideally. Might want to consider using aenter and aexit if this is async

docker-compose.yml
```diff
@@ -54,5 +54,20 @@ services:
       - db_data:/var/lib/mysql
     command: [ 'mysqld', '--character-set-server=utf8mb4', '--collation-server=utf8mb4_unicode_ci' ]
 
+  weaviate:
```
asim-shrestha, line 57: might need to set this up in unit tests aswell

PR #708, opened 2023-06-05
🥳 Improve max loops configuration

awtkns:
- Removed all max loop logic from JS
- Max loops are now controlled by python settings
- Ported a large part of auth code
- Local user now must sign in (fear not an error message is shown)

TODO:
- docs / cli improvement?

### DIFF
--- a/platform/reworkd_platform/schemas.py
+++ b/platform/reworkd_platform/schemas.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from typing import Optional, List, Literal
 
 from pydantic import BaseModel, Field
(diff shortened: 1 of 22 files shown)


platform/reworkd_platform/schemas.py
```diff
@@ -17,14 +25,47 @@ class ModelSettings(BaseModel):
...
+
+class AgentTaskAnalyze(AgentRun):
+    task: str
+    tool_names: List[str] = Field(default=[], alias="toolNames")
+    model_settings: ModelSettings = Field(default=ModelSettings())
+
+
+class AgentTaskExecute(AgentRun):
+    task: str
+    analysis: Optional[Analysis] = None  # TODO Why is this optional?
```
asim-shrestha, line 47: Was optional because using analyses / tools was introduced later. We should use it

PR #724, opened 2023-06-06
📈 The start of token tracking

awtkns:


platform/reworkd_platform/services/tokenizer/dependencies.py
```diff
@@ -0,0 +1,7 @@
+from fastapi import Request
+
+from reworkd_platform.services.tokenizer.service import TokenService
```
asim-shrestha, line 3: TokenizerService?

platform/reworkd_platform/services/tokenizer/lifetime.py
```diff
@@ -0,0 +1,16 @@
+import tiktoken
+from fastapi import FastAPI
+
+ENCODING_NAME = "cl100k_base"  # gpt-4, gpt-3.5-turbo, text-embedding-ada-002
```
asim-shrestha, line 4: Will need some mechanism to change this based on ModelSettings in the future.

platform/reworkd_platform/services/tokenizer/service.py
```diff
@@ -0,0 +1,22 @@
...
+        return self.encoding.encode(text)
+
+    def detokenize(self, tokens: list[int]) -> str:
+        return self.encoding.decode(tokens)
+
+    def token_count(self, text: str) -> int:
+        return len(self.tokenize(text))
+
+    def get_context_space(
+        self, prompt: str, max_tokens: int, reserved_response_tokens: int
```
asim-shrestha, line 18: The model object itself may have a max_tokens defined, not sure

asim-shrestha (approved):

platform/reworkd_platform/services/tokenizer/lifetime.py
awtkns: I think we can look at this in the future, right now tiktoken only seems to support openai

PR #770, opened 2023-06-12
📄 Separate settings page

asim-shrestha:


next/src/utils/interfaces.ts
```diff
@@ -1,15 +1,29 @@
...
+  temperature: number;
+  max_tokens: number;
+}
+
+export const toApiModelSettings = (modelSettings: ModelSettings) => ({
+  language: modelSettings.language.name,
+  model: modelSettings.customModelName,
+  temperature: modelSettings.customTemperature,
+  max_tokens: modelSettings.maxTokens,
+});
```
awtkns, line 16: Might be better to package this in the agent api module

awtkns (approved):

PR #793, opened 2023-06-13
feat(#747): place colour theming into tailwind config

### LINKED ISSUE #747: 💅 Place colour theming into tailwind config
Tailwind suggests saving theme colours into the config file as evidenced in this document: https://tailwindcss.com/docs/customizing-colors

We should do this in a way such that a given colour will preserve light and dark info

Jshen123:
Resolves #747

**Description**: 
add color theming for dark mode (see [figma](https://www.figma.com/file/hBQdHjGlnESMcHXcaPn1lM/AgentGPT?type=design&node-id=168-2&t=b7JXtYxDqGd4HViO-0) for detail - ignore coloring for light)

### DIFF
--- a/next/tailwind.config.cjs
+++ b/next/tailwind.config.cjs
@@ -64,16 +64,6 @@ module.exports = {
           }
         }
       },
-      display: {
-        100: {
-          dark: "#C0C0C0",
-        },
-      },
-      body: {
-        100: {
-          dark: "#FFFFFF"
-        }
-      }
     },
   },
   plugins: [
--- a/next/tailwind.config.cjs
+++ b/next/tailwind.config.cjs
@@ -37,7 +37,7 @@ module.exports = {
             dark: "#ff9f0a",
           },
         },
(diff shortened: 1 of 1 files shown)


next/tailwind.config.cjs
```diff
@@ -16,8 +16,65 @@ module.exports = {
...
+        }
+      },
+      display: {
+        100: {
+          dark: "#C0C0C0",
+        },
+      },
+      body: {
+        100: {
+          dark: "#FFFFFF"
```
asim-shrestha, line 74: What are these for?

asim-shrestha (approved):

next/tailwind.config.cjs
Jshen123: they were supposed to be for text (display for headings and body for body texts)

**I removed them**. Current thought process is to streamline this and use colors under shades/grays for text colors  but I also think we have more flexibility long-term if we decouple text colors and grays

PR #808, opened 2023-06-15
🎨Advanced Settings Updates

awtkns:
- All users should now be able to see settings


asim-shrestha:
image

I don't think users will be able to tell why these settings don't work. 

- We need to maybe darken colours on input
- We need to change the cursor to 🚫 on hover
- We can maybe say sign in to change or something

asim-shrestha:
Also, I think we should use a different colour for the border, or maybe a hr between the sections

PR #809, opened 2023-06-15
🐛Sidebar improvements

awtkns:
- Sidebar now stays open/closed on reload
- Sidebar doesn't flicker between page changes


asim-shrestha:
Some notes

- I think best practice for this would be that the sidebar is handled purely through CSS. Have some ideas for how we can accomplish this
- We should have the sidebar open by default. There's a lot of information users might miss if it isn't open on desktop
- I'm not sure we should save this state between sessions. For reasons above, I think it should always open on refresh

PR #846, opened 2023-06-21
🥷 Extract tasks from messages in autonomous agent

asim-shrestha:


next/src/components/console/ChatWindowTitle.tsx
```diff
@@ -0,0 +1,30 @@
+import type { GPTModelNames } from "../../types";
+import { GPT_35_TURBO_16K, GPT_4 } from "../../types";
+import React from "react";
+
+export const ChatWindowTitle = ({ model }: { model: GPTModelNames }) => {
```
awtkns, line 5: This could be cleaner in the future, maybe a dict of models to spans?

next/src/services/agent/agent-run-model.tsx
```diff
@@ -11,21 +11,27 @@ export interface AgentRunModel {
 
   getRemainingTasks(): Task[];
 
+  getCurrentTask(): Task | undefined;
+
```
awtkns, line 15: not sure why there are new lines here?

next/src/types/message.ts
```diff
@@ -29,3 +29,14 @@ export const nonTaskScehma = z
...
+/*
+ * Ideal message type
+ * {
+ *  icon: IconType,
+ *  title: string,
+ *  subtitle: string, // Optional
+ *  value: string, // Markdown formatted value
+ *  color: string, // Classname used for the border
+ * }
+ */
```
awtkns, line 42:
```suggestion
```

awtkns (approved):

next/src/types/message.ts
awtkns: actually you can leave the above

next/src/components/console/ChatWindowTitle.tsx
asim-shrestha: Will add a ticket

next/src/services/agent/agent-run-model.tsx
asim-shrestha: Formatter just doing this, not sure

next/src/types/message.ts
asim-shrestha: Yeah, hopefully we can use this in the future

Issue #850, opened 2023-06-23
🐛 Errors when running docker-compose build


yoventures:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/reworkd/AgentGPT/labels/bug) didn't find any similar reports.

### Expected Behavior

when running docker-compose build, the building process should complete successfully.

### Current behaviour

<img width="850" alt="Screenshot 2023-06-22 201706" src="https://github.com/reworkd/AgentGPT/assets/114469342/7795f473-e952-4459-8c7f-09abccd9fda4">

Getting the following error when running **docker-compose build**

`#0 31.48   ChefBuildError
#0 31.48
#0 31.48   Backend subprocess exited when trying to invoke get_requires_for_build_wheel
#0 31.48
#0 31.48   /bin/sh: 1: pkg-config: not found
#0 31.48   /bin/sh: 1: pkg-config: not found
#0 31.48   Trying pkg-config --exists mysqlclient
#0 31.48   Command 'pkg-config --exists mysqlclient' returned non-zero exit status 127.
#0 31.48   Trying pkg-config --exists mariadb
#0 31.48   Command 'pkg-config --exists mariadb' returned non-zero exit status 127.
#0 31.48   Traceback (most recent call last):
#0 31.48     File "/usr/local/lib/python3.11/site-packages/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
#0 31.48       main()
#0 31.48     File "/usr/local/lib/python3.11/site-packages/pyproject_hooks/_in_process/_in_process.py", line 335, in main
#0 31.48       json_out['return_val'] = hook(**hook_input['kwargs'])
#0 31.48     File "<string>", line 27, in find_package_name
#0 31.48   Exception: Can not find valid pkg-config name.
#0 31.48   Specify MYSQLCLIENT_CFLAGS and MYSQLCLIENT_LDFLAGS env vars manually
#0 31.48
#0 31.48
#0 31.48   at /usr/local/lib/python3.11/site-packages/poetry/installation/chef.py:152 in _prepare
#0 31.49       148│
#0 31.49       149│                 error = ChefBuildError("\n\n".join(message_parts))
#0 31.49       150│
#0 31.49       151│             if error is not None:
#0 31.49     → 152│                 raise error from None
#0 31.49       153│
#0 31.49       154│             return path
#0 31.49       155│
[...]

Issue #848, opened 2023-06-22
Start script fails in openSuse 15.5


gcpdev:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/reworkd/AgentGPT/labels/bug) didn't find any similar reports.

### Expected Behavior

First of all, thanks for this great application, I used the web version and it seems to achieve my goals way better than other (more famous) agents.

When following the Readme instructions and running:
`git clone https://github.com/reworkd/AgentGPT.git && cd AgentGPT
./setup.sh`
should start the application

### Current behaviour

On my machine, it unfortunately fails. Full log as follows:
```
# ./setup.sh

up to date, audited 146 packages in 659ms

27 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

> agentgpt-cli@1.0.0 start
> node src/index.js

 █████╗  ██████╗ ███████╗███╗   ██╗████████╗ ██████╗ ██████╗ ████████╗
██╔══██╗██╔════╝ ██╔════╝████╗  ██║╚══██╔══╝██╔════╝ ██╔══██╗╚══██╔══╝
███████║██║  ███╗█████╗  ██╔██╗ ██║   ██║   ██║  ███╗██████╔╝   ██║   
██╔══██║██║   ██║██╔══╝  ██║╚██╗██║   ██║   ██║   ██║██╔═══╝    ██║   
██║  ██║╚██████╔╝███████╗██║ ╚████║   ██║   ╚██████╔╝██║        ██║   
╚═╝  ╚═╝ ╚═════╝ ╚══════╝╚═╝  ╚═══╝   ╚═╝    ╚═════╝ ╚═╝        ╚═╝   
                                                                      
Welcome to the AgentGPT CLI! This CLI will generate the required .env files.
Copies of the generated envs will be created in `./next/.env` and `./platform/.env`.

Existing ./next/env file found. Validating...
? How will you be running AgentGPT? 🐋 Docker-compose (Recommended)
[+] Building 52.0s (32/35)                                                                                          
 => [agentgpt-next internal] load build definition from Dockerfile                                             0.0s
 => => transferring dockerfile: 32B                                                                            0.0s
 => [mysql:8.0 internal] load build definition from Dockerfile                                                 0.0s
 => => transferring dockerfile: 31B                                                                            0.0s
 => [agentgpt-platform internal] load build definition from Dockerfile                                         0.1s
[...]

PR #852, opened 2023-06-23
pkg-config: not found when `poetry install --only main`;

chennqqi:
Ref:
https://github.com/reworkd/AgentGPT/issues/848
https://github.com/reworkd/AgentGPT/issues/850


gcpdev:
I can confirm this fixes the issue.

hutao562:
> I can confirm this fixes the issue.

me too

PR #853, opened 2023-06-23
Fix Dockerfile + missing env vars in example

tim-codes:
Just finished configuring the project locally and had to make a few corrections to get it working:
1. Platform Dockerfile was failing for missing pkg-config dependency
2. .env file required 2 more max-loops vars to startup properly

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,7 @@
+# Platform Config: (max loops determines how many times the agent may execute)
+NEXT_PUBLIC_MAX_LOOPS="100"
+REWORKD_PLATFORM_MAX_LOOPS="100"
+
 # Deployment Environment:
 NODE_ENV="development"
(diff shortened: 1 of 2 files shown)


asim-shrestha:
Thanks! Lets do the following instead

```
REWORKD_PLATFORM_MAX_LOOPS=${NEXT_PUBLIC_MAX_LOOPS}
```

Also note, if you run the setup commands (`./setup.sh` for mac) It should tell you when you have an out of date ENV value

asim-shrestha:
Checking in @tim-codes :)

tim-codes:
Hey @asim-shrestha thanks will update today :)

tim-codes:
@asim-shrestha changes made. FYI I did do the setup.sh script originally, but still needed the env var amendments.

asim-shrestha:
Thank you @tim-codes!

PR #864, opened 2023-06-26
Ergomake

vieiralucas:


docker-compose.yml
```diff
@@ -5,14 +5,21 @@ services:
     container_name: next
     build:
       context: ./next
-      dockerfile: Dockerfile
+      args:
+        NEXT_PUBLIC_BACKEND_URL: 'http://localhost:8000'
+        NEXTAUTH_URL: "http://localhost:3000"
+        DATABASE_URL: "mysql://reworkd_platform:reworkd_platform@db:3307/reworkd_platform"
+        NEXTAUTH_SECRET: "changeme"
```
asim-shrestha, line 12: Is there any way to simply pass an env into the ergomake env?

docker-compose.yml
```diff
@@ -22,10 +29,6 @@ services:
     ports:
       - "8000:8000"
     restart: always
-    volumes:
-      - ./platform:/app/src/
-    env_file:
-      - next/.env
```
asim-shrestha, line 28: How come we're removing this?

docs/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM node:19-alpine
```
asim-shrestha, line 1: Nice! Needed this :)

next/Dockerfile
```diff
@@ -5,6 +5,18 @@ ARG NODE_ENV
...
+ENV NEXT_PUBLIC_BACKEND_URL=$NEXT_PUBLIC_BACKEND_URL
+
+ARG NEXTAUTH_URL
+ENV NEXTAUTH_URL=$NEXTAUTH_URL
+
+ARG DATABASE_URL
+ENV DATABASE_URL=$DATABASE_URL
+
+ARG NEXTAUTH_SECRET
+ENV NEXTAUTH_SECRET=$NEXTAUTH_SECRET
```
asim-shrestha, line 18: Ideally we don't have this. It introduces another vector that we need to update whenever a new env variable is introduced. It would be cool to have a Vercel style workflow of adding ENV values to a project

next/Dockerfile
```diff
@@ -29,11 +41,12 @@ COPY entrypoint.sh /
...
 
+RUN npm run build
 
 # Expose the port the app will run on
 EXPOSE 3000
 
 ENTRYPOINT ["sh", "/entrypoint.sh"]
 
 # Start the application
-CMD ["npm", "run", "dev"]
```
asim-shrestha, line 39: Ideally we don't build here. I personally use this for dev and require hot reloads :)

asim-shrestha:
Thanks @vieiralucas! Left some comments - looking forward to getting this up and running!

vieiralucas:
@asim-shrestha I moved every customization to inside the `.ergomake` folder.

The thing about `.env` is that it is not commited into the source, so it works great when you're running locally, but `ergomake` can not build the compose when it performs a "fresh" clone.

Running things in dev mode is problematic because it usually needs too much resources and/or reduces the performance too much.
Finally, since every `preview` has its own link, we need a mechanism to customize the `backend` url that the frontend uses, also customize the origins the backend accept `cors`, etc.

So, that is why I made some changes to the compose and dockerfiles.

That said, I completely agree that these things should not interfere on how developers wants to runs things on their machines. So, better to just move things to `.ergomake`.

asim-shrestha:
Thanks @vieiralucas!

> @asim-shrestha I moved every customization to inside the `.ergomake` folder.

Perfect :)

> Running things in dev mode is problematic because it usually needs too much resources and/or reduces the performance too much.

Ah, makes sense

> The thing about `.env` is that it is not committed into the source, so it works great when you're running locally, but `ergomake` can't build the compose when it performs a "fresh" clone.
> Finally, since every `preview` has its own link, we need a mechanism to customize the `backend` url that the frontend uses, also customize the origins the backend accept `cors`, etc.
> So, for these reasons I made some changes to the compose and dockerfiles.

I think the changes such as the mappings of env to the ergomake variables make sense given this, the ideal for us is that we'd be able to create our own `.ergomake.env` or something that we store in your cloud (Or manage on the web with something like Vercel's ENV editor) 

We'll try to review and get this in later today!

awtkns (approved):

awtkns:
Looks good to me! I like how it is clean and separated! I think we can merge then iterate should the need arise.

PR #866, opened 2023-06-26
Update .env.example

Darkhorse12022023:


Darkhorse12022023:
[38e9449](https://github.com/reworkd/AgentGPT/commit/38e9449ba94c46446f0c46c6d51cd081e6964351)

Darkhorse12022023:
[38e9449](https://github.com/reworkd/AgentGPT/commit/38e9449ba94c46446f0c46c6d51cd081e6964351)

asim-shrestha:
Hey @Darkhorse12022023, I think we'll have to elect to keep `changeme` as one shouldn't be using the same value as the example env anyways. Thank you for the help though!

Issue #750, opened 2023-06-09
🌚 Add light mode options to components


asim-shrestha:
Currently, all of our components only have a default option that is dark flavoured. We should change these to the "dark" setting and also add light colours.

PR #881, opened 2023-06-28
feat(#750): add themestore and usetheme

Jshen123:


next/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  const preferredThemeMatches = (
+    matchObj || window?.matchMedia(`(prefers-color-scheme: ${PREFERRED_THEME})`)
+  ).matches;
+
+  return theme === PREFERRED_THEME || (theme === "system" && preferredThemeMatches);
+}
+
+export function handleTheme(theme, matchObj?) {
```
asim-shrestha, line 17: Can we use arrow syntax. Also, should we have type params for these functions

next/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,41 @@
+import { type Theme, THEMES } from "../types";
+import { useEffect } from "react";
+const PREFERRED_THEME = "dark"; // preferred theme must be dark for Tailwind
```
asim-shrestha, line 3: How come?

next/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,41 @@
...
+  } else {
+    classList.remove(PREFERRED_THEME);
+  }
+}
+
+export function useTheme(theme: Theme) {
+  theme = THEMES.includes(theme) ? theme : "system";
+
+  useEffect(() => {
+    const preferredTheme = window.matchMedia(`(prefers-color-scheme: ${PREFERRED_THEME})`);
```
asim-shrestha, line 35: What's going on here? What if their preferred theme isn't dark?

next/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,41 @@
...
+    return;
+  }
+
+  const classList = document.documentElement.classList;
+
+  if (isPreferredTheme(theme, matchObj)) {
+    classList.add(PREFERRED_THEME);
+  } else {
+    classList.remove(PREFERRED_THEME);
+  }
```
asim-shrestha, line 28: Whats going on here, worth some comments?

next/src/stores/themeStore.ts
```diff
@@ -0,0 +1,57 @@
...
+        name: "theme",
+        storage: createJSONStorage(() => localStorage),
+        onRehydrateStorage: () => {
+          return (state, error) => {
+            if (error) {
+              console.error("an error happened during hydration. ", error);
+            } else {
+              handleTheme(state ? state.theme : "system");
+            }
+          };
```
asim-shrestha, line 50: Neat, so this way even if we have local storage we won't run into rehydration issues with SSR? If so, could also use these in the other stores (Had to use a separate hook for them to ensure there were no issues)

PR #882, opened 2023-06-28
🐛 Fix Negative Token Sizing

### LINKED ISSUE #878: reworkd_platform.web.api.errors.OpenAIError: -351 is less than the minimum of 1 - 'max_tokens'
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/reworkd/AgentGPT/labels/bug) didn't find any similar reports.

### Expected Behavior

I encoutered this issue when running the docker. 
Thanks,
Daniel
image

### Current behaviour

doesn't seem to produce any results

### Steps to reproduce

run the containter and put in a task

### Possible solution

I don't know

### Which Operating Systems are you using?

- [ ] Android
- [ ] iPhone/iPad
- [ ] Linux
- [X] macOS
- [ ] Windows

### Acknowledgements

- [X] My issue title is concise, descriptive, and in title casing.
- [X] I have searched the existing issues to make sure this bug has not been reported yet.
- [X] I am using the latest version of AgentGPT.
- [X] I have provided enough information for the maintainers to reproduce and diagnose the issue.

awtkns:
fixes #878 - Was able to reporduce the issue by having a very small max token number. We now ensure that there is a minimum of 200 tokens as prompt return space.

### DIFF
--- a/platform/reworkd_platform/web/api/agent/agent_service/open_ai_agent_service.py
+++ b/platform/reworkd_platform/web/api/agent/agent_service/open_ai_agent_service.py
@@ -1,7 +1,6 @@
 from typing import List, Optional
 
 from lanarky.responses import StreamingResponse
-
 from langchain.callbacks.base import AsyncCallbackHandler
 from langchain.output_parsers import PydanticOutputParser
 from langchain.prompts import ChatPromptTemplate, SystemMessagePromptTemplate
@@ -35,6 +34,8 @@ from reworkd_platform.web.api.memory.memory import AgentMemory
 
 
 class OpenAIAgentService(AgentService):
+    MIN_COMPLETION_SPACE = 200
+
     def __init__(
         self,
         model: WrappedChatOpenAI,
@@ -54,7 +55,7 @@ class OpenAIAgentService(AgentService):
             [SystemMessagePromptTemplate(prompt=start_goal_prompt)]
         )
 
-        self.model.max_tokens -= self.token_service.count(
+        self.calculate_max_tokens(
             prompt.format_prompt(
                 goal=goal,
                 language=self.language,
@@ -89,10 +90,10 @@ class OpenAIAgentService(AgentService):
             language=self.language,
         )
 
-        self.model.max_tokens -= self.token_service.count(prompt.to_string())
-
-        # TODO: We are over counting tokens here
-        self.model.max_tokens -= self.token_service.count(str(functions))
+        self.calculate_max_tokens(
+            prompt.to_string(),
+            str(functions),
+        )
 
         message = await openai_error_handler(
             func=self.model.apredict_messages,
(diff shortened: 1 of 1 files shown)


platform/reworkd_platform/web/api/agent/agent_service/open_ai_agent_service.py
```diff
@@ -177,3 +176,8 @@ async def create_tasks_agent(
                 memory.add_tasks(unique_tasks)
 
         return unique_tasks
+
+    def calculate_max_tokens(self, *args: str):
+        prompt_tokens = sum([self.token_service.count(arg) for arg in args])
+        self.model.max_tokens -= prompt_tokens
+        self.model.max_tokens = max(self.model.max_tokens, self.MIN_COMPLETION_SPACE)
```
asim-shrestha, line 183: So the issue is that we require the following to hold true
```
requested_tokens + prompt_tokens <= model_max_tokens
```

Our current solution performs the following:
```
requested_tokens - prompt_tokens 
```
Where `requested_tokens` <= `model_max_tokens`. This becomes a problem when we have a low requested_tokens value as the calculation can lead to negative numbers. 

Furthermore, suppose a user requested 100 tokens but the prompt itself took 99 tokens. The user would then only see a 1 token response when they themselves would expect 100 tokens. A bit contrived of an example, but a bad experience nonetheless. 

Not only that, but imagine we have a 4000 token prompt, and then we try to use a min completion space of 200. This will cause an error as 4200 is too large.

We should instead do the following:

```
requested_tokens <= model_max_tokens - prompt_tokens
```

Which basically means 

```
self.model.max_tokens = min(self.model.max_tokens, get_model_max_tokens({"gpt-3.5", "gpt-4", etc}) - prompt_tokens)
```

platform/reworkd_platform/web/api/agent/agent_service/open_ai_agent_service.py
```diff
@@ -177,3 +177,10 @@ async def create_tasks_agent(
                 memory.add_tasks(unique_tasks)
 
         return unique_tasks
+
+    def calculate_max_tokens(self, *prompts: str) -> None:
+        model_max_tokens = LLM_MODEL_MAX_TOKENS.get(self.model.model_name, 4000)
+        prompt_tokens = sum([self.token_service.count(p) for p in prompts])
+        requested_tokens = model_max_tokens - prompt_tokens
```
asim-shrestha, line 184: should be called max_allowed_tokens or something

platform/reworkd_platform/web/api/agent/model_settings.py
```diff
@@ -28,17 +30,9 @@ def create_model(
...
         model_kwargs={"user": user.email},
     )
-
-
-def get_model_name(model_str: LLM_Model) -> str:
-    if model_str == "gpt-4":
-        return "gpt-4-0613"
-    if model_str == "gpt-3.5-turbo":
-        return "gpt-3.5-turbo-0613"
-    return model_str
```
asim-shrestha, line 44: Nice, yeah they updated to always use this stuff now

platform/reworkd_platform/web/api/agent/agent_service/open_ai_agent_service.py
```diff
@@ -35,6 +35,8 @@
 
 
 class OpenAIAgentService(AgentService):
+    MIN_COMPLETION_SPACE = 200
```
asim-shrestha, line 38: delete

Issue #876, opened 2023-06-27
🔑 Issues with Serper 


asim-shrestha:
It seems many people are using the incorrect Serp key. They use google.serper or something instead of serper.dev. The CLI already includes a link to serper.dev.

We should think to update our documentation website to include a link to serper as well and explicitly mention it doesn't work with Google serper.

MusabShakeel576:
@asim-shrestha please assign this issue to me, I'll update the docs.

asim-shrestha:
Thanks @MusabShakeel576! Merged and left a comment in the PR!

MusabShakeel576:
@asim-shrestha I looked at your comment. We already have an interactive setup that guides users on which API is needed, and we also have a .env file that lists all the required APIs. Wouldn't it be repetitive to mention the same thing again in a new API section in the Setup Docs?

jasangill1:
Hello, @MusabShakeel576 Greate question !!! If we have all these things and users are still getting confused there must be a reason, by adding as much clarity as possible we can try negating these errors!!!

MusabShakeel576:
@jasangill1, should we improve the onboarding process by throwing proper errors and warnings when the user doesn't have AgentGPT properly configured, or should we update the docs and add an API section?

asim-shrestha:
@MusabShakeel576 maybe you're right, we could instead have a warning message about how we don't support the other serp apis lets start with this.

Alternatively, the setup.sh script itself could validate these keys for the users as the write them. This should be handled in a different issue however

jasangill1:
Hello, @MusabShakeel576 Amazing suggestion!!! I have created an issue that I can assign you if you would like to tackle these issues, that would be greatly appreciated. 😊

Feel free to message me whenever you need help !!

jasangill1:
Sounds good @MusabShakeel576

PR #883, opened 2023-06-28
docs: Update SERP API

MusabShakeel576:
Include Serper.dev and explicity mention that other SERP APIs are not supported.

Issue: #876


asim-shrestha:
Thanks @MusabShakeel576! Perhaps also worth mentioning in https://docs.reworkd.ai/development/setup. Maybe we can make an API keys section before the ENV section. We should also put the links to get API keys for openai, replicate, serper, etc

PR #885, opened 2023-06-28
748 add scroll to bottom button

abhithory:
Hello,

- Scroll to the bottom button created
- Created a HideShow component for adding hiding and showing fade animation to the button 
- used `animate-bounce` for bobbling animation from Tailwind


asim-shrestha:
Thanks @abhithory! 

Some updates I made:
- Used existing state to keep track of scroll. 
- Refactored function to be reused

abhithory:
@asim-shrestha Nice.....

One thing I want to mention....

- with the existing `hasUserScrolled` state, then the button will be shown immediately.....
but I think in the issue you mentioned that the button should show after the user scrolled far enough...

image

PR #897, opened 2023-06-29
🎉 Workflows

awtkns:
When deployed, we will turn off this page so that others can access it till it is finished. This pr finishes the basic CRUD structure for workflows.

In the UI you can now CREATE / VIEW a list of your workflows. When you click into a workflow you can then edit the workflow by adding nodes and edges. These changes are persisted in the database on save. Some followup things that will need to be addressed later:

# TODO in followup:
- WORKFLOW NODES SHOULD BE CREATED AT CENTER OF VIEWPORT
  - Currently they are made at 0,0 (which is problematic if the user is elsewhere)

- Different node type

- Cycles in the graph (e.g. A -> B -> C -> A) should be prevented against

- Loading
  - Mention unsaved changes
  
- Different node types
  - e.g. "Start" and "End" nodes
  - Blocks / integrations

- Delete entire workflows


asim-shrestha (changes requested):
- [ ] Fix mypy
- [ ] Fix node tests

next/prisma/schema.prisma
```diff
@@ -130,3 +130,48 @@ model Task {
   @@index([run_id])
   @@map("agent_task")
 }
+
+// WORKFLOW
+model Workflow {
+  id              String    @id @default(cuid())
+  user_id         String
```
asim-shrestha, line 137: Are we going to have foreign keys?

next/prisma/schema.prisma
```diff
@@ -130,3 +130,48 @@ model Task {
...
+  organization_id String?
+  name            String
+  description     String    @default("")
+  create_date     DateTime  @default(now())
+  update_date     DateTime? @updatedAt
+  delete_date     DateTime?
```
asim-shrestha, line 143: Will we keep history of changes? Usually these workflow tools save revisions that you can go back and forth from. Of course we can add later, but we should probably think about this now

How to do this per chat GPT
https://chat.openai.com/share/01ba1f78-371b-4ebc-bddd-32b88e4770f7

next/src/components/workflow/BasicNode.tsx
```diff
@@ -0,0 +1,48 @@
...
+        "border-translucent rounded-md bg-white p-3 shadow-xl shadow-gray-300 dark:border-neutral-900 dark:shadow-2xl dark:shadow-black",
+        "dark:bg-stone-900 dark:text-white dark:shadow-stone-800"
+      )}
+    >
+      <div className="flex items-center">
+        {/*{block?.image && (*/}
+        {/*  <div>*/}
+        {/*    <Image src={block.image} alt={"Test"} width={30} height={30} />*/}
+        {/*  </div>*/}
+        {/*)}*/}
```
asim-shrestha, line 22: Can we remove these if we're not using them?

next/src/hooks/useMounted.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const useMounted = () => {
+  const [mounted, setMounted] = useState(false);
+
+  useEffect(() => {
+    setMounted(true);
+  }, []);
+
+  return mounted;
+};
```
asim-shrestha, line 11: What are we using this for?

next/src/services/fetch-utils.ts
```diff
@@ -0,0 +1,68 @@
+import { env } from "../env/client.mjs";
+import type { z } from "zod";
+
+export const get = async <T extends z.ZodTypeAny>(
```
asim-shrestha, line 4: Why not Axios instead of fetch?

next/src/services/api-utils.ts
```diff
@@ -1,7 +1,6 @@
 import axios from "axios";
 import type { Session } from "next-auth";
 import { env } from "../env/client.mjs";
-import type { z } from "zod";
 
 export const post = async <T>(url: string, body: unknown, session?: Session) => {
```
asim-shrestha, line 5: Whats the difference in using this axios based post instead of the fetch based post?

next/src/components/workflow/Flowchart.tsx
```diff
@@ -0,0 +1,125 @@
...
+const FlowChart: FC<FlowChartProps> = ({
+  isLoading,
+  onSave,
+  // workflow,
+  nodesModel,
+  edgesModel,
+  ...props
+}) => {
+  const mounted = useMounted();
+  // const { theme } = useTheme();
```
asim-shrestha, line 47: Remove

next/src/components/workflow/Flowchart.tsx
```diff
@@ -0,0 +1,125 @@
...
+      {/*      /!*<WorkflowRunDialog workflow={workflow} />*!/*/}
+      {/*    </div>*/}
+      {/*    <div className="flex items-center gap-2 text-sm font-thin dark:text-white">*/}
+      {/*      <p>Last saved: {new Date().toLocaleDateString()}</p>*/}
+      {/*      <span>*/}
+      {/*        <IoIosCheckmarkCircleOutline color="green" size={20} />*/}
+      {/*      </span>*/}
+      {/*    </div>*/}
+      {/*  </div>*/}
+      {/*)}*/}
```
asim-shrestha, line 93: Del

next/prisma/schema.prisma
awtkns: No fk's. Planetscale doesn't support fks

next/prisma/schema.prisma
awtkns: I think it good to think about, however IMHO this is over-engineering for now. Lets get workflows working and people using then built this out

next/src/pages/workflow/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+          }}
+        >
+          New Workflow
+        </Button>
+      </div>
+    </SidebarLayout>
+  );
+};
+
+export default WorkflowList;
```
asim-shrestha, line 57: Missing translation static props. Same with other page i imagine

next/src/services/api-utils.ts
awtkns: Nothing, was trying out fetch after seeing some comparisons. I think we can just stick to axios. It is simple

PR #898, opened 2023-06-29
🚨 Improved error logging

awtkns:
- We no longer trigger sentry errors for errors caused by peoples own api keys.
- No more sleepless nights :)

### DIFF
--- a/platform/reworkd_platform/web/api/agent/helpers.py
+++ b/platform/reworkd_platform/web/api/agent/helpers.py
@@ -6,6 +6,7 @@ from langchain.schema import BaseOutputParser, OutputParserException
 from openai import InvalidRequestError
 from openai.error import AuthenticationError, RateLimitError, ServiceUnavailableError
 
+from reworkd_platform.schemas import ModelSettings
 from reworkd_platform.web.api.errors import OpenAIError
 
 T = TypeVar("T")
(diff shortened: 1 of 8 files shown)


platform/reworkd_platform/web/api/agent/helpers.py
```diff
@@ -30,24 +31,28 @@ async def openai_error_handler(
             e,
             "OpenAI is experiencing issues. Visit "
             "https://status.openai.com/ for more info.",
+            should_log=not settings.custom_api_key,
```
asim-shrestha, line 34: we need to wrap the model and also pass this in

platform/reworkd_platform/web/api/agent/model_settings.py
```diff
@@ -17,14 +17,11 @@ class WrappedChatOpenAI(ChatOpenAI):
...
 ) -> WrappedChatOpenAI:
-    if model_settings.custom_api_key != "":
-        api_key = model_settings.custom_api_key
-    else:
-        api_key = rotate_keys(
-            gpt_3_key=settings.openai_api_key,
-            gpt_4_key=settings.secondary_openai_api_key,
-            model=model_settings.model,
-        )
+    api_key = model_settings.custom_api_key or rotate_keys(
```
asim-shrestha, line 20: What if the frontend sends ""?

awtkns: "" evaluates to false in python (and also js)

platform/reworkd_platform/web/api/agent/helpers.py
awtkns: Can you elaborate?

PR #899, opened 2023-06-29
Fix Agent-API retries

istandre:
## Changes
Per the comment written in the withRetries function, the onError function returns whether to continue. Thus, if true, we should not break out, but continue looping instead.

I also removed the second condition since it is redundant (already handled by the for loop).

Finally, the for loop is now starting with i=0, since, per my understanding, retries should be the number of retries to do after an error and not the number of tries in total (number of tries = 1 + number of retries).

## Testing
I have added a test file to test the withRetries function.

Note that I have added a mock file for the window.matchMedia function to avoid polluting the tests logs.
Without the mock, even if no test fails, we see the following error in the logs:
```shell
console.error
   an error happened during hydration.  TypeError: window?.matchMedia is not a function
[...]
```

Note also that I have added "test" as possible values for the NEXT_PUBLIC_VERCEL_ENV env var. Even with the SKIP_ENV_VALIDATION env var enabled, the client config gets loaded and parsed via agent-api.ts (import api-utils.ts -> import client.mjs)


asim-shrestha:
Wow thanks @istandre! Would never expect people to tackle something like this, especially without an existing ticket none-the-less! 

I've gone ahead and fixed up some minor stuff. Seems like the mock you included was due to the file import having a direct dependency on some `stores` which can't run unless they're on the browser. I've gone ahead and refactored it a bit. Also changed up some naming and ensured that the `work` was returning true if it should continue retrying and false otherwise (Was backwards)

Please let me know if you want to work on anything else and I'm happy to set you up with a ticket you're excited for. I'm Asim#3933 on the discord or you can email me at asim@reworkd.ai

PR #914, opened 2023-06-30
🦅 Header title sytling 

jasangill1:
## About
Update title text to match landing page design

## Notes
- Let me know if I should change the inline styling to use tailwind instead 
- Didn't add satoshi variable font as it looks different from the one in Figma 

Screen Shot 2023-06-30 at 12 32 37 PM

### DIFF
--- a/next/src/components/landing/Hero.tsx
+++ b/next/src/components/landing/Hero.tsx
@@ -10,7 +10,7 @@ const Hero = () => {
   const router = useRouter();
 
   return (
-    <div className="grid grid-cols-1 gap-4 sm:grid-cols-2">
+    <div className="grid grid-cols-1 gap-2 sm:grid-cols-2">
       <div className="col-span-1">
         <FadeIn duration={1.5} delay={0}>
           <div className="mb-2">
@@ -27,13 +27,15 @@ const Hero = () => {
           </div>
           <h1
             className={clsx(
-              "text-md text-left leading-none tracking-[-0.64px] sm:text-xl sm:leading-[1.1em]",
+              "text-4xl sm:text-4xl md:text-6xl lg:text-7xl",
+              "text-left",
+              "pb-2",
               "bg-clip-text text-transparent",
-              "via-[rgba(255, 255, 255, 0.22)] bg-gradient-to-br from-white to-neutral-500",
-              "text-xl font-bold leading-[68px] tracking-[-0.64px] sm:text-[60px] sm:leading-[1.1em]"
+              "bg-gradient-to-br from-white to-neutral-600",
+              "leading-[1.1em] tracking-[-0.5px]"
             )}
           >
-            <div className="pb-3">
+            <div>
               Autonomous AI
               <br />
               Agents At Your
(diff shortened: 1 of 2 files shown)


next/src/components/landing/Hero.tsx
```diff
@@ -5,80 +5,69 @@ import PrimaryButton from "../PrimaryButton";
...
+              backgroundImage:
+                "linear-gradient(172.22deg, #FFFFFF 13.42%, rgba(255, 255, 255, 0.22) 154.02%)",
+              fontFamily: "",
+              fontSize: "68px",
+              fontWeight: 700,
+              lineHeight: "68px",
+              letterSpacing: "0px",
+              textAlign: "left",
+            }}
+          >
```
asim-shrestha, line 44: All of this styling should be done with tailwind if possible. Gradients are sometimes hard, but fontFamily, fontSize, fontWeight, everything else should definitely be Tailwind

next/src/components/landing/Hero.tsx
```diff
@@ -5,80 +5,69 @@ import PrimaryButton from "../PrimaryButton";
...
-            "bg-gradient-to-br from-white via-neutral-300 to-neutral-500 bg-clip-text text-transparent",
-            "text-center text-4xl font-bold leading-[42px] tracking-[-0.64px] sm:text-6xl sm:leading-[1.1em]"
-          )}
-        >
-          <div>
-            Autonomous AI Agents
-            <br />
-            at Your Fingertips
+    <div className="grid grid-cols-1 gap-4 sm:grid-cols-2">
+      <div className="col-span-1 mt-8">
```
asim-shrestha, line 14: Random `mt-8`. Lets get rid of this

next/src/components/landing/Hero.tsx
```diff
@@ -5,80 +5,64 @@ import PrimaryButton from "../PrimaryButton";
...
+            className={clsx(
+              "text-md text-left leading-none tracking-[-0.64px] sm:text-xl sm:leading-[1.1em]",
+              "bg-clip-text text-transparent",
+              "bg-gradient-to-br from-white via-neutral-300 to-neutral-500",
+              "text-xl font-bold leading-[68px] tracking-[-0.64px] sm:text-[60px] sm:leading-[1.1em]"
+            )}
+            style={{
+              backgroundImage:
+                "linear-gradient(172.22deg, #FFFFFF 13.42%, rgba(255, 255, 255, 0.22) 154.02%)",
+            }}
```
asim-shrestha, line 38: The tailwind styling already has a gradient. Use that as a reference for how it will work

asim-shrestha: We don't need inline styles at all here

next/src/components/landing/Hero.tsx
```diff
@@ -5,80 +5,62 @@ import PrimaryButton from "../PrimaryButton";
...
-          onClick={() => {
-            router.push("/").catch(console.error);
-          }}
-        >
-          Get started
-        </PrimaryButton>
-      </FadeIn>
+          <h1
+            className={clsx(
+              "text-4xl sm:text-4xl md:text-6xl lg:text-7xl",
```
asim-shrestha, line 30: nit: You can remove `sm:text-4xl` since its already `text-4xl` by default

PR #918, opened 2023-06-30
🧊 3D image added

jasangill1:
## About the PR

- Added our new bg image 
- Adjusting the current gradient to match the new image
- Getting positioning correct

## Notes 
- Let me know if the gradient needs adjustment if it should be bright or darker if so at what side
 -Was having trouble with x overflow but got it working!

- Design decision
 
 -Left the backing as is will adjust in the next PR

Screen Shot 2023-06-30 at 3 34 17 PM


next/src/pages/secret-landing-page.tsx
```diff
@@ -27,6 +27,7 @@ const HomePage = () => {
           </div>
         </div>
       </div>
+      <Image src="/3DBG.png" alt="Background Image" layout="fill" quality={100} />
```
asim-shrestha, line 30: Let's call it hero background

next/src/pages/secret-landing-page.tsx
```diff
@@ -27,6 +27,7 @@ const HomePage = () => {
           </div>
         </div>
       </div>
+      <Image src="/heroBackground.png" alt="Background Image" layout="fill" quality={100} />
```
asim-shrestha, line 30: nit: `hero-background`

Issue #905, opened 2023-06-30
🦅 Navbar Component for the landing page 


jasangill1:
## Build out a navbar component:

- **Navbar Component**: The first component we need to construct is a navigation bar (navbar). This will be the primary means for users to navigate through our website.

- **Design**: The navbar design should follow the structure shown in the provided images. Please ensure the styling matches the images as closely as possible.

- **Light and Dark Mode Toggle**: Include a button for toggling between light and dark mode on the navbar. For this stage of development, we don't need to implement the functionality behind the light and dark mode switch. Focus on the aesthetics and placement of this button.

- **Responsiveness**: Please ensure that the navbar is responsive and maintains its design integrity across various screen sizes and devices.

- **Links**: Attach all links accordingly 

REMINDER: Remember to use the in-app components like buttons, inputs, combo box, and dialog box wherever need be these can be found in the /ui file.

![Navbar Design](https://github.com/reworkd/AgentGPT/assets/56135840/d83861dd-9239-4348-9997-a5a970cb6851)

awtkns:
Just to add to this the nav bar should not be sticky

abhithory:
@jasangill1, I want to work on this... can you please assign me this issue?

jasangill1:
Created this nav bar but now changing to a newly designed nav bar

PR #923, opened 2023-07-01
🎨 Gradient Update

awtkns:
- Took a break from workflows because i wanted to play with gradients


awtkns:
# Before
Screenshot 2023-07-01 at 11 18 58 AM

# After
Screenshot 2023-07-01 at 11 16 45 AM

next/src/layout/sidebar.tsx
```diff
@@ -90,7 +90,7 @@ const SidebarLayout = (props: PropsWithChildren) => {
                     </button>
                   </div>
                   <FadingHr className="my-2" />
-                  <div className="mb-2 flex-1 overflow-y-auto">
+                  <div className="-mr-2 mb-2 flex-1 overflow-y-auto">
```
asim-shrestha, line 93: Why negative margin?

awtkns: got carried away make this nice:
**
image
**
vs 
image

awtkns:
Nav bar border removed and no longer made fixed. Related to #905

next/src/layout/sidebar.tsx
asim-shrestha: Should we update other margins instead of introducing negatives? Feels like this (and stuff like negative translations) might be harder to maintain

PR #933, opened 2023-07-02
💪 Fix mypy <> sqla typing issues

awtkns:
- Upgrading to use mapped_collumn instead worked like a charm. This is the standard for SLQA2.0. Prior we were using 1.4 syntax. Thanks chat gpt.

### DIFF
--- a/platform/pyproject.toml
+++ b/platform/pyproject.toml
@@ -17,7 +17,7 @@ uvicorn = { version = "^0.22.0", extras = ["standard"] }
 pydantic = { version = "^1.10.8", extras = ["dotenv"] }
 yarl = "^1.9.2"
 ujson = "^5.7.0"
-SQLAlchemy = { version = "^2.0.15", extras = ["asyncio"] }
+sqlalchemy = {version = "^2.0.17", extras = ["mypy", "asyncio"]}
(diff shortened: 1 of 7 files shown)


platform/pyproject.toml
```diff
@@ -64,6 +64,8 @@ multi_line_output = 3
 src_paths = ["reworkd_platform", ]
 
 [tool.mypy]
+# https://docs.sqlalchemy.org/en/20/orm/extensions/mypy.html TODO: enable this if tying issues happen with SQLA
+# plugins = ["sqlalchemy.ext.mypy.plugin"]
```
awtkns, line 68:
```suggestion
```

platform/reworkd_platform/db/models/user.py
```diff
@@ -29,18 +23,13 @@ class UserSession(Base):
 class User(Base):
     __tablename__ = "User"
 
-    id = Column(
-        String,
-        primary_key=True,
-        default=lambda _: str(uuid.uuid4()),
-        unique=True,
-        nullable=False,
```
asim-shrestha, line 37: Missing id

asim-shrestha (approved):

PR #937, opened 2023-07-03
🐛 Fix bug that arises when using negative max_tokens

asim-shrestha:
## About
We can request a negative max token value when the completion tokens are that significant

This results in a strange response from OpenAI: 
```
-2473 is less than the minimum of 1 - 'max_tokens'
```

This PR will ensure we have a minimum value of 1 `max_tokens`. This will result in the following response:

```
This model's maximum context length is 4097 tokens. However, your messages resulted in 6478 tokens. Please reduce the length of the messages.
```

Also added some tests for this method and moved it into the token_service


platform/reworkd_platform/services/tokenizer/token_service.py
```diff
@@ -0,0 +1,26 @@
+from tiktoken import Encoding
+
+from reworkd_platform.schemas import LLM_MODEL_MAX_TOKENS
+from reworkd_platform.web.api.agent.model_settings import WrappedChatOpenAI
+
+
+class TokenService:
```
awtkns, line 7: Nice!

awtkns (approved):

awtkns:
Love the tests ❤️

Issue #938, opened 2023-07-03
👟 Footer section for new landing page


awtkns:
<img width="1143" alt="Screenshot 2023-07-02 at 6 19 24 PM" src="https://github.com/reworkd/AgentGPT/assets/32209255/3b661f4e-db00-45fe-bbd2-ced3bf43547a">

PR #940, opened 2023-07-03
👟 Footer for landing page

awtkns:
relates to #938 

Text alignment isn't quite perfect but this can be addressed later

### DIFF
--- a/next/src/components/landing/Footer.tsx
+++ b/next/src/components/landing/Footer.tsx
@@ -1,0 +1,87 @@
+import React from "react";
+import Image from "next/image";
+import { FaDiscord, FaGithub, FaLinkedin, FaTwitter } from "react-icons/fa";
+import clsx from "clsx";
+
+interface Props {
+  className?: string;
+}
+
+const links = [
+  {
+    name: "Github",
+    href: "https://github.com/reworkd/AgentGPT",
+    icon: FaGithub,
+  },
+  {
+    name: "Twitter",
+    href: "https://twitter.com/ReworkdAI",
+    icon: FaTwitter,
+  },
+  {
+    name: "Discord",
+    href: "https://discord.gg/gcmNyAAFfV",
+    icon: FaDiscord,
+  },
+  {
+    name: "LinkedIn",
+    href: "https://www.linkedin.com/company/reworkd/",
+    icon: FaLinkedin,
+  },
+];
+
+const Footer = (props: Props) => {
+  return (
+    <footer className={clsx(props.className, "relative w-full text-sm font-extralight")}>
+      <div className="ml-auto mr-auto w-full max-w-7xl px-8 pb-16 text-center">
+        <div className="flex w-full flex-col items-baseline justify-between gap-16 sm:flex-row sm:justify-start sm:gap-0">
+          <div className="flex font-extralight">
+            <Image src="wordmark.svg" width="200" height="20" alt="Reworkd AI" className="mr-2" />
+          </div>
+          <div className="flex flex-1 justify-center">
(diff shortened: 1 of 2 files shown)


next/src/components/landing/Footer.tsx
```diff
@@ -0,0 +1,89 @@
...
+    name: "Discord",
+    href: "https://discord.gg/gcmNyAAFfV",
+    icon: FaDiscord,
+  },
+  {
+    name: "LinkedIn",
+    href: "https://www.linkedin.com/company/reworkd/",
+    icon: FaLinkedin,
+  },
+];
```
asim-shrestha, line 31: `links.tsx` has this. Lets reuse that - we can also add the LinkedIn link there

next/src/components/landing/Footer.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const Footer = (props: Props) => {
+  return (
+    <footer className={clsx(props.className, "relative w-full text-sm font-extralight")}>
+      <div className="ml-auto mr-auto w-full max-w-7xl px-8 pb-16 text-center">
+        <div className="flex w-full flex-col items-baseline justify-between gap-16 sm:flex-row sm:justify-start sm:gap-0">
+          <div className="flex font-extralight">
```
asim-shrestha, line 38:
```suggestion
          <div className="flex">
```

next/src/components/landing/Footer.tsx
```diff
@@ -0,0 +1,89 @@
...
+                href={link.href}
+                target="_blank"
+                rel="noopener noreferrer"
+                className="transition-all hover:rotate-6 hover:text-purple-400"
+              >
+                <span className="sr-only">{link.name}</span>
+                <link.icon size={24} />
+              </a>
+            ))}
+          </div>
```
asim-shrestha, line 78: What is this one for? Its just always hidden?

next/src/components/landing/Footer.tsx
```diff
@@ -0,0 +1,89 @@
...
+            <Image src="wordmark.svg" width="200" height="20" alt="Reworkd AI" className="mr-2" />
+          </div>
+          <div className="flex flex-1 justify-center">
+            <div className="flex flex-col gap-6 px-8 text-center text-gray-400 sm:flex-row sm:text-left md:tracking-wider">
+              <a className="text-center" href="https://reworkd.ai/privacyPolicy" target="_blank">
+                Privacy Policy
+              </a>
+              <a className="text-center" href="http://reworkd.ai/terms">
+                Terms and Conditions
+              </a>
```
asim-shrestha, line 48: Could also place these in the links section. Also, one of these is missing a target="_blank"

next/src/components/landing/Footer.tsx
awtkns: I saw that but I want to use the IconType not the component here

next/src/components/landing/Footer.tsx
awtkns: This one is shows up on large screens, notice the `sm:flex`

asim-shrestha:
image

On mobile:
- Icons are too large
- Logo / text should be centered
- Privacy policy text, etc button text should be larger

In general:
- Privacy policy text, etc button text should have a hover highlight

next/src/components/landing/Footer.tsx
awtkns: Ideally links.tsx should use the icontype aswell. but that can be future

next/src/components/landing/Footer.tsx
asim-shrestha: Ah mb, but we could just add selector styling for the gap and do just two link icons instead of doing two of the entire thing?

PR #942, opened 2023-07-03
✨ Summary button

asim-shrestha:
## About
- Adds a button to summarize the run after a run has "stopped"

## Notes
- Add a new backend route and function in open ai agent service for summary
- Given we don't have sources in this summary, create a new summary prompt
- Force GPT-3.5-16k for this summary to ensure it is as long as possible
- NOTE even with the prompting, it will still make up information :D


next/src/components/console/ChatWindow.tsx
```diff
@@ -133,4 +137,43 @@ const ChatWindow = ({
...
+    .tasks()
+    .filter((task) => task.status == "completed" && task.result !== "");
+  const [summarized, setSummarized] = useState(false);
+
+  // Reset the summarized state when the agent changes
+  useEffect(() => {
+    setSummarized(false);
+  }, [agent]);
+
+  console.log(tasksWithResults);
```
awtkns, line 153: remove

next/src/components/console/ChatWindow.tsx
```diff
@@ -133,4 +137,43 @@ const ChatWindow = ({
   );
 };
 
+const Summarize = () => {
```
awtkns, line 140: nit: Ideally instead of this being a component, it would be cool if this could just be a different type of message. I can see a world where we may need human input in a message, etc...

platform/reworkd_platform/web/api/agent/agent_service/open_ai_agent_service.py
```diff
@@ -181,3 +183,24 @@ async def create_tasks_agent(
...
+    ) -> FastAPIStreamingResponse:
+        self.model.model_name = "gpt-3.5-turbo-16k"
+        self.model.max_tokens = 8000  # Total tokens = prompt tokens + completion tokens
+
+        snippet_max_tokens = 7000  # Leave room for the rest of the prompt
+        text_tokens = self.token_service.tokenize("".join(results))
+        text = self.token_service.detokenize(text_tokens[0:snippet_max_tokens])
+        logger.info(f"Summarizing text: {text}")
+
+        return summarize(
```
awtkns, line 201: I think we should also have error handling here

platform/reworkd_platform/web/api/agent/tools/search.py
```diff
@@ -81,4 +84,4 @@ async def call(
         if len(snippets) == 0:
             return stream_string("No good Google Search Result was found", True)
 
-        return summarize(self.model, self.language, goal, task, snippets)
+        return summarize_with_sources(self.model, self.language, goal, task, snippets)
```
awtkns, line 87: should this be async? does this need to be awaited?

platform/reworkd_platform/web/api/agent/dependancies.py
```diff
@@ -103,3 +104,10 @@ async def agent_create_validator(
     crud: AgentCRUD = Depends(agent_crud),
 ) -> AgentTaskCreate:
     return await validate(body, crud, "create")
+
+
+async def agent_summarize_validator(
```
awtkns, line 109: nice!

awtkns (approved):
Looks good to me! Just minor stuff :)

next/src/components/console/ChatWindow.tsx
asim-shrestha: Would require a larger refactor of how messages are displayed + would need to be decoupled from tasks.

I agree though

platform/reworkd_platform/web/api/agent/tools/search.py
asim-shrestha: This is a streaming response. Seems the async stuff is handled automatically through Lanarky

platform/reworkd_platform/web/api/agent/agent_service/open_ai_agent_service.py
asim-shrestha: Ah true, should do this across the board for tools / streaming responses

Issue #892, opened 2023-06-28
🗝️ Validation for User Keys in setup.sh script


jasangill1:
### 🔖 Feature description

During the setup process, users are currently responsible for the correctness of the keys they enter. However, this leaves room for human error and can lead to unsuccessful setups due to invalid keys.

### ✔️ Solution

Integrating a validation step for the keys within the setup.sh script itself. This step would automatically verify the accuracy of the keys as users input them, reducing the risk of errors and subsequent setup issues.

This solution would involve:

1. Implementing a key validation function within the setup.sh script.
2. Testing the new function to ensure its effectiveness and reliability.
3. Updating the setup documentation to inform users of this new feature.

By integrating a key validation step directly into the setup process, we can significantly improve the user experience and minimize setup-related issues.

MusabShakeel576:
@jasangill1, instead of adding validation logic in setup.sh, it would be better to leverage already existing AgentGPT CLI and add `validate` methods similar to OpenAIApiKey in both [serpApiKey and replicateApiKey objects](https://github.com/reworkd/AgentGPT/blob/83589430b470c29751f1f42a62f2d0860608ac93/cli/src/questions/newEnvQuestions.js#L19-L30).

jasangill1:
Hello @MusabShakeel576 Sorry if my instructions were unclear but that was the intention would you like to tackle this issue I will assign it to you if so!!!

Feel free to reach out if you com across any troubles !!!

MusabShakeel576:
@jasangill1 no worries. Sure, I'll look into it.

jasangill1:
Hello @MusabShakeel576 Feel free to message me whenever you need help!!

MusabShakeel576:
@jasangill1 pushed, and please answer the question I mentioned in the second point in PR description.

jasangill1:
Hello @MusabShakeel576 Sounds good ill get @asim-shrestha to review!!

jasangill1:
Great work @MusabShakeel576 going to close this issue as you have completed it!!!

PR #957, opened 2023-07-03
feat: Validate user API keys in CLI

MusabShakeel576:
This solution involves:
- [x]  Implement a key validation function within the setup.sh/newEnvQuestions.js.
- [x] Test the new function to ensure its effectiveness and reliability. (There is no testing tool setup in the `cli` folder, I did manual testing, should I set up one or does manual testing suffice for now?)
- [x] Update the setup documentation to inform users of this new feature. (Already mentioned in [Interactive Setup](https://docs.reworkd.ai/development/setup#interactive-setup))

Issue: #892

### DIFF
--- a/cli/src/helpers.js
+++ b/cli/src/helpers.js
@@ -20,9 +20,7 @@ export const printTitle = () => {
 
 // Function to check if entered api key is in the correct format or empty
 export const isValidKey = (apikey, pattern) => {
-  if (pattern.test(apikey) || apikey === "") {
-    return true;
-  } else {
-    return "\nInvalid api key. Please try again.";
-  }
+  return (apikey === "" || pattern.test(apikey))
 };
+
+export const validKeyErrorMessage = "\nInvalid api key. Please try again."
--- a/cli/src/helpers.js
+++ b/cli/src/helpers.js
@@ -18,8 +18,11 @@ export const printTitle = () => {
   );
 };
 
-// Function to check if entered api key is in the correct format
-export const isValidSkKey = (apikey) => {
-  const pattern = /^sk-[a-zA-Z0-9]{48}$/;
-  return pattern.test(apikey);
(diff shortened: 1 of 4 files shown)


asim-shrestha:
Thanks @MusabShakeel576! 

As per testing, I was thinking we could even make an example call to their endpoints here to see if if works. This should suffice for now though as long as the google serp key is flagged by this!

asim-shrestha:
Had to update the replicate validator to the following:

`return isValidKey(apikey, /^r8_[a-zA-Z0-9]{37}$/)`

So `r8_` instead of `r8-`. Do you have a key with `r8-`?

cli/src/helpers.js
```diff
@@ -18,8 +18,11 @@ export const printTitle = () => {
   );
 };
 
-// Function to check if entered api key is in the correct format
-export const isValidSkKey = (apikey) => {
-  const pattern = /^sk-[a-zA-Z0-9]{48}$/;
-  return pattern.test(apikey);
+// Function to check if entered api key is in the correct format or empty
+export const isValidKey = (apikey, pattern) => {
+  if (pattern.test(apikey) || apikey === "") {
```
awtkns, line 23:
```suggestion
  if (apikey === "" || pattern.test(apikey)) {
```

awtkns (approved):
looks good to me once Asim's changes are addressed

MusabShakeel576:
@asim-shrestha added example API calls, and update Replicate key.

cli/src/helpers.js
MusabShakeel576: @awtkns that's a nice catch to prevent extra computation, I fixed it.

asim-shrestha:
Awesome @MusabShakeel576! 

Made the following change to fix some issues:
`🐛 Add node-fetch and allow api keys to be empty strings`

Thanks for the help here! This will be really valuable for new devs

PR #963, opened 2023-07-04
✨ Fix bug and remove unused combobox

asim-shrestha:


next/src/components/Input.tsx
```diff
@@ -162,8 +131,8 @@ const Input = (props: InputProps) => {
   return (
     <div
       className={clsx(
-        `items-left z-5 text-color-primary flex h-fit w-full flex-col rounded-xl font-mono text-lg md:flex-row md:items-center`,
-        `md:flex-row md:items-center`
+        "items-left z-5 text-color-primary flex h-fit w-full flex-col rounded-xl font-mono text-lg md:flex-row md:items-center",
+        "md:flex-row md:items-center"
```
awtkns, line 135: nit: Probably best to not use clsx here as then the imports dont get sorted. Instead the editor should word wrap

awtkns (approved):

PR #976, opened 2023-07-05
901: Reorder Tasks in Task Queue

abhithory:
Hello,

- Added Reorder Tasks using Drag and Drap
- Here I have implemented custom logic for updating the state because the default `Reorder` feature from framer-motion continuously updates the state when we drag and drop the element and this was hanging the website. (because we have to update the global state)
- So, I have implemented a function to update the global state only when we drag over another element.

https://github.com/reworkd/AgentGPT/assets/76877003/84da0a0b-bc10-43f3-b9e5-24e071543a26

e we have to update the global state)
- So, I have implemented a function to update the global state only when we drag over another element.

Thanks,
Abhishek

### DIFF
--- a/next/src/components/TaskWindow.tsx
+++ b/next/src/components/TaskWindow.tsx
@@ -9,7 +9,7 @@ import clsx from "clsx";
 import Input from "./Input";
 import Button from "./Button";
 import { v1 } from "uuid";
-import { AnimatePresence } from "framer-motion";
+import { AnimatePresence, Reorder } from "framer-motion";
 import { MESSAGE_TYPE_TASK, Task, TASK_STATUS_STARTED } from "../types/task";
 import { useTaskStore } from "../stores/taskStore";
 
@@ -21,6 +21,7 @@ export const TaskWindow = ({ visibleOnMobile }: TaskWindowProps) => {
   const [customTask, setCustomTask] = React.useState("");
   const agent = useAgentStore.use.agent();
   const tasks = useTaskStore.use.tasks();
+  const setTasks = useTaskStore.use.setTasks();
   const addTask = useTaskStore.use.addTask();
   const [t] = useTranslation();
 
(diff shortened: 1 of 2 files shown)


next/src/components/TaskWindow.tsx
```diff
@@ -53,11 +87,28 @@ export const TaskWindow = ({ visibleOnMobile }: TaskWindowProps) => {
               This window will display agent tasks as they are created.
             </p>
           )}
-          <AnimatePresence>
-            {tasks.map((task, i) => (
-              <Task key={i} index={i} task={task} />
-            ))}
-          </AnimatePresence>
+          <Reorder.Group drag="y" values={[]} onReorder={() => {}} as="div" >
```
asim-shrestha, line 90: How come we have no values and no onReorder? 

It seems like you've manually built reorder capabilities but we can leverage framer motion here to do it all :)

asim-shrestha:
Thanks @abhithory, left some comments. I think we can let framer handle the complexities of drag and drop.

Additionally, we have lost some spacing between the tasks

image

asim-shrestha:
Lets also add a drag trigger to the bottom left
https://www.framer.com/motion/reorder/##drag-triggers

image

next/src/components/TaskWindow.tsx
abhithory: hello @asim-shrestha, I have tried leveraging the framer motion for reordering the Tasks but 
- it continuously updates the Tasks list when we drag and drop and we are changing the global state, so the website got the hang
- also in default reorder framer motion, we can rearrange tasks above completed tasks because there is no option to not drag and drop tasks above the completed task

so, that's why I added custom logic for drag and drop.

In custom drag and drop:-
- we only update the task's global state only on dragging over a new element, this will not continually update and the website will not hang
- also we can stop it drag tasks to the above-completed tasks

abhithory:
> Thanks @abhithory, left some comments. I think we can let framer handle the complexities of drag and drop.
> 
> Additionally, we have lost some spacing between the tasks
> 
> image

abhithory:
@

abhithory:
hello @asim-shrestha, 
- I can handle complete drag and drop with framer motion, it is simple, and I can do
 but I have mentioned the reason why I added the custom logic.

please read that and let me know what should I do?

abhithory:
hi @asim-shrestha, i have added rearrange with the default framer motion in this PR #996 .
please check this. 

but this is not working properly because of above mentioned reasons.

abhithory:
hi @asim-shrestha , 
- i have added spacing and a drag trigger button. please check it

PR #977, opened 2023-07-06
⚡️ Chat with Agent

asim-shrestha:
## About
Adds functionality to chat with an agent run

## Info
- Each chat session will **ONLY** get the results of the previous tasks as context. It will not get the goal. It will not get the task names. It won't even get previous chat messages. This is only to modify the results of the chat
- It will use the same "loop" limit as the types of messages. This means pro users will get 25 chat messages. 
- Chatting will trigger a pause so that the agent isn't doing a whole bunch of other shit while the chat is in progress
- Had to place `ChatControls` inside ChatWindow instead of just in the `index` so that scroll would behave as expected.


platform/reworkd_platform/web/api/agent/views.py
```diff
@@ -94,6 +96,19 @@ async def summarize(
     )
 
 
+@router.post("/chat")
```
awtkns, line 99: This may be confusing with our enterprise `/chat` endpoint.

platform/reworkd_platform/web/api/agent/prompts.py
```diff
@@ -122,3 +122,16 @@
     """,
     input_variables=["goal", "language", "query", "snippets"],
 )
+
+chat_prompt = PromptTemplate(
```
awtkns, line 126: Nice!

platform/reworkd_platform/web/api/agent/dependancies.py
```diff
@@ -111,3 +114,10 @@ async def agent_summarize_validator(
     crud: AgentCRUD = Depends(agent_crud),
 ) -> AgentSummarize:
     return await validate(body, crud, "summarize")
+
+
+async def agent_chat_validator(
+    body: AgentChat = Body(),
+    crud: AgentCRUD = Depends(agent_crud),
+) -> AgentChat:
+    return await validate(body, crud, "chat")
```
awtkns, line 123: This will allow people to call this 30 times per agent. It should be fine, but something to consider in the prod version

platform/reworkd_platform/web/api/agent/agent_service/open_ai_agent_service.py
```diff
@@ -204,3 +207,38 @@ async def summarize_task_agent(
...
+            prompt.format_prompt(
+                language=self.settings.language,
+            ).to_string(),
+        )
+
+        print(
+            prompt.format_prompt(
+                language=self.settings.language,
+            ).to_string()
+        )
```
awtkns, line 237:
```suggestion
```

platform/Dockerfile
```diff
@@ -4,6 +4,7 @@ RUN apt-get update && apt-get install -y \
   default-libmysqlclient-dev \
   pkg-config \
   gcc \
+  ca-certificates \
```
awtkns, line 7: whats this for?

next/src/utils/interfaces.ts
```diff
@@ -22,6 +22,7 @@ export const toApiModelSettings = (modelSettings: ModelSettings, session?: Sessi
 };
 
 export interface RequestBody {
+  run_id?: string;
```
awtkns, line 25: We should split this interface up eventually like we did for the backend (future)

awtkns (approved):

platform/Dockerfile
asim-shrestha: Heard that might fix my issues with langchain. Seems to be cert related or some crap

platform/reworkd_platform/web/api/agent/agent_service/open_ai_agent_service.py
asim-shrestha: Will delete

platform/reworkd_platform/web/api/agent/dependancies.py
asim-shrestha: Yeah, once we move to token / call based it will be better

platform/reworkd_platform/web/api/agent/views.py
asim-shrestha: true, should be fine for now

Issue #984, opened 2023-07-06
failed to solve: failed to read dockerfile


VectorZhao:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/reworkd/AgentGPT/labels/bug) didn't find any similar reports.

### Expected Behavior

run docker

### Current behaviour

(base) ➜  AgentGPT git:(main) ./setup.sh --docker

up to date, audited 154 packages in 609ms

30 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

> agentgpt-cli@1.0.0 start
> node src/index.js

 █████╗  ██████╗ ███████╗███╗   ██╗████████╗ ██████╗ ██████╗ ████████╗
██╔══██╗██╔════╝ ██╔════╝████╗  ██║╚══██╔══╝██╔════╝ ██╔══██╗╚══██╔══╝
███████║██║  ███╗█████╗  ██╔██╗ ██║   ██║   ██║  ███╗██████╔╝   ██║   
██╔══██║██║   ██║██╔══╝  ██║╚██╗██║   ██║   ██║   ██║██╔═══╝    ██║   
██║  ██║╚██████╔╝███████╗██║ ╚████║   ██║   ╚██████╔╝██║        ██║   
╚═╝  ╚═╝ ╚═════╝ ╚══════╝╚═╝  ╚═══╝   ╚═╝    ╚═════╝ ╚═╝        ╚═╝   
                                                                      
Welcome to the AgentGPT CLI! This CLI will generate the required .env files.
Copies of the generated envs will be created in `./next/.env` and `./platform/.env`.

Existing ./next/env file found. Validating...
? How will you be running AgentGPT? 🐋 Docker-compose (Recommended)
[+] Building 0.1s (8/8) FINISHED                                                                                                                                                                
 => [agentgpt-docs internal] load build definition from Dockerfile                                                                                                                         0.0s
 => => transferring dockerfile: 2B                                                                                                                                                         0.0s
 => [agentgpt-docs internal] load .dockerignore                                                                                                                                            0.1s
 => => transferring context: 2B                                                                                                                                                            0.0s
 => CANCELED [agentgpt-platform internal] load .dockerignore                                                                                                                               0.1s
 => CANCELED [agentgpt-platform internal] load build definition from Dockerfile                                                                                                            0.1s
 => CANCELED [mysql:8.0 internal] load build definition from Dockerfile                                                                                                                    0.0s
[...]

PR #987, opened 2023-07-06
Fix/failed to read dockerfile/issue 984

Gsciarra:
This is to fix [this issue](https://github.com/reworkd/AgentGPT/issues/984).
A `services` `docs` was added, but the `Dockerfile` is not present in the `docs` directory.


asim-shrestha:
Hey thanks! We just need to include the dockerfile again

PR #989, opened 2023-07-06
fix ergomake

vieiralucas:
I'm not familiar with `minlify`.

I don't know if it is possible to build and serve a production version outside of their platform, I did not find out how to do that.

For now, this removes `docs` from `ergomake`.

Similar to what was done at #987


asim-shrestha:
Thanks @vieiralucas!

Issue #888, opened 2023-06-28
🪵 Loss of Agent Data upon Login


jasangill1:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [X] I searched previous [Ideas in Discussions](https://github.com/reworkd/AgentGPT/discussions/categories/ideas) didn't find any similar feature requests.
- [X] I searched previous [Issues](https://github.com/reworkd/AgentGPT/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

Currently, when a user clicks on "deploy agent" without being logged in, they are prompted to log in. However, after successfully logging in, the agent data entered before login is not preserved. This forces the user to re-type all the agent data, which can be a cumbersome process and negatively affect user experience.

### ✔️ Solution

To solve this issue, we propose implementing a feature to cache the user's input data temporarily. This could be achieved by storing the agent data in the local storage or a temporary session. Once the user successfully logs in, the stored data can be retrieved and populated in the appropriate fields automatically, thereby preserving the user's work and improving the overall user experience.

### Acknowledgements

- [X] My issue title is concise, descriptive, and in title casing.
- [X] I have searched the existing issues to make sure this feature has not been requested yet.
- [X] I have provided enough information for the maintainers to understand and evaluate this request.

jasangill1:
Merged !!!

PR #991, opened 2023-07-08
✨ Fixed Loss of Agent Data upon Login

jasangill1:
## About

- added the functionality to store agent data in the browser's local storage and retrieve it when returning to the login home page after login.

## Technical Decisions:

- Chosen to use the browser's localStorage API due to its simplicity and synchronicity.

https://github.com/reworkd/AgentGPT/issues/888

https://github.com/reworkd/AgentGPT/assets/56135840/792a85bd-daba-4928-9f9f-33fd947159c7

### DIFF
--- a/next/src/pages/index.tsx
+++ b/next/src/pages/index.tsx
@@ -92,6 +92,7 @@ const Home: NextPage = () => {
 
   const handleNewAgent = (name: string, goal: string) => {
     if (session === null) {
+      storeAgentDataInLocalStorage(name, goal);
       setShowSignInDialog(true);
       return;
     }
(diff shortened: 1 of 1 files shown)


next/src/pages/index.tsx
```diff
@@ -121,6 +122,28 @@ const Home: NextPage = () => {
...
+    if (session !== null) {
+      const agentData = getAgentDataFromLocalStorage();
+
+      if (agentData) {
+        setNameInput(agentData.name);
+        setGoalInput(agentData.goal);
+        localStorage.removeItem("agentData");
+      }
+    }
+  }, [session]);
```
asim-shrestha, line 145: Ideally we let this be done through a call back mechanism that would provide these values as query parameters. This solution seems ok for now, will merge tomorrow

PR #992, opened 2023-07-08
🚀 Workflow Execution Engine

awtkns:


awtkns:
Any thoughts @asim-shrestha @jasangill1?

asim-shrestha (approved):

next/src/env/schema.mjs
```diff
@@ -1,5 +1,5 @@
 // @ts-check
-import { z } from "zod";
+import {z} from "zod";
```
asim-shrestha, line 2: Might have some incorrect `on save` in your public branch

next/src/hooks/useSocket.ts
jasangill1, line 1: What is the scketi used for ?

PR #1004, opened 2023-07-09
💬 1003 update hero section text

jasangill1:
## About 

- Changed the weight spacing tracking leading and got rid of icons on banner besides arrow.


next/src/components/BannerBadge.tsx
```diff
@@ -18,7 +17,6 @@ const BannerBadge = ({ children, className, ...props }: BadgeProps) => {
       )}
       {...props}
     >
-      <IoSparkles className="mx-1" />
```
asim-shrestha, line 21: Oh we can probably keep this

PR #1005, opened 2023-07-09
🐼 Updated the gradient

jasangill1:
## About 

- Added clsx 
- Adjusted the text to be aligned on the edges by using rem
- Made the text normal from thin (kinda liked the thin font let me know what you think)

image

### DIFF
--- a/next/src/components/landing/Hero.tsx
+++ b/next/src/components/landing/Hero.tsx
@@ -23,7 +23,7 @@ const Hero = () => {
           </div>
           <h1
             className={clsx(
-              "pb-2 text-left text-2xl font-normal leading-loose tracking-wide sm:text-6xl md:text-7xl",
+              "pb-2 text-left text-2xl font-normal leading-loose tracking-[.09rem] sm:text-6xl md:text-7xl",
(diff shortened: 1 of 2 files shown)


next/src/components/landing/Hero.tsx
```diff
@@ -34,7 +36,10 @@ const Hero = () => {
           </h1>
           <p
             className={clsx(
-              "text-16 font-inter my-3 mb-9 inline-block w-full text-left align-top font-thin leading-[28px] tracking-widest"
+              "font-inter my-3 mb-9 inline-block w-full",
```
asim-shrestha, line 39: instead of defining mb-9 here, we can let the outer container deal with spacing

next/src/components/landing/Hero.tsx
```diff
@@ -34,7 +36,10 @@ const Hero = () => {
...
+              "text-left align-top font-light leading-[28px] tracking-[.08rem]",
```
asim-shrestha, line 40: How come we're manually doing leading / tracking? Are the defaults not good enough?

Issue #1007, opened 2023-07-10
Update backing.tsx


Jshen123:
![Image](https://user-images.githubusercontent.com/50181239/252166413-c75070a1-504b-4ee2-85eb-e5b128b33fae.png)

PR #1010, opened 2023-07-10
feat(#1007): implement backing component

Jshen123:
image


next/next.config.mjs
```diff
@@ -19,6 +19,11 @@ const config = {
       poll: 1000,
       aggregateTimeout: 300
     };
+    config.module.rules.push({
```
Jshen123, line 22: leverage SVGR (to transform svgs into react components)
https://react-svgr.com/

asim-shrestha: Whats the motivation for SVGR? Can't we just load them as SVG?

PR #1018, opened 2023-07-11
✨ Small workflow updates

asim-shrestha:
Splitting up for a smaller PR

- Update right sidebar to be dynamic
- Add basics of the workflow node logic
- Add frontend to display available workflow nodes: Still need to make it clickable, actually do stuff


docs/development/workflows.mdx
```diff
@@ -0,0 +1,20 @@
+---
```
awtkns, line 1: Nice!

awtkns (approved):

next/public/tools/web.png
awtkns, line 1: wonder if FA icons can be used instead

platform/reworkd_platform/web/api/workflow/blocks/web/status_check.py
```diff
@@ -0,0 +1,31 @@
...
+    description = "Outputs the status code of a GET request to a URL"
+    image_url = ""
+    input_config: UrlStatusCheckBlockInput
+
+    def run(self) -> BlockIOBase:
+        try:
+            response = requests.get(self.input_config.url)
+            code = response.status_code
+        except RequestException:
+            code = None
```
awtkns, line 28: might be good to log something?

next/public/tools/web.png
asim-shrestha: Not sure we'll have applicable icons for all images in the future

platform/reworkd_platform/web/api/workflow/blocks/web/status_check.py
asim-shrestha: Yeah we need better logging throughout

PR #1021, opened 2023-07-11
fix: apply anti-aliasing to text

Jshen123:
Without anti-aliasing
image

WIth anti-aliasing
image


awtkns (approved):

awtkns:
LGTM! Waiting for @asim-shrestha

asim-shrestha:
So what does this do exactly? Does it affect performance?

PR #1034, opened 2023-07-13
➡️ 1027 navbar top right buttons

jasangill1:
##About 

- Used the primary buttons as a template,

- Font styling not apply to button wondering if this was the correct approach

- Still working on screen alignment for nav layout, Looking for review.

image


next/src/components/NavLayout.tsx
```diff
@@ -13,10 +16,15 @@ const navigation = [
...
 export default function NavLayout({ children }: { children: ReactNode }) {
+  const router = useRouter();
+
   return (
     <>
       <FadeIn duration={3}>
-        <Disclosure as="nav" className="absolute top-0 z-50 w-full bg-transparent text-white">
+        <Disclosure
+          as="nav"
+          className="hidden:overlay absolute top-0 z-50 w-full bg-transparent text-white"
```
awtkns, line 26: Ideally we want to be using absolute as less as possible as this will cause problems down the road with small screens. I think we can do the header - content - footer all with flex

next/src/components/NavbarTextButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+    <Button
+      onClick={onClick}
+      className={clsx(
+        "flex h-8 items-center justify-center rounded-full border-[1px] border-black font-light",
+        "transition duration-200 ease-in-out hover:bg-neutral-900 hover:text-neutral-100 focus-visible:bg-neutral-900 focus-visible:outline-none focus-visible:ring-4 focus-visible:ring-neutral-500",
+        "bg-transparent text-white/60"
+      )}
+    >
+      {icon}
+      {children}
```
awtkns, line 21: Maybe we should just make this children and pass everything in for max configurability :)

next/src/pages/landing-page.tsx
```diff
@@ -7,7 +7,7 @@ const HomePage = () => {
     <NavLayout>
       <div className="flex w-full justify-center">
         <div className="flex max-w-screen-2xl flex-col items-center justify-center overflow-x-hidden px-5 text-white">
-          <div className="flex h-screen w-full flex-col items-start justify-center overflow-x-hidden px-4 text-white lg:pl-16">
```
asim-shrestha, line 10: lets keep this. Good to have some padding

next/src/components/NavbarContactButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+      className={clsx(
+        "flex h-8 items-center justify-center rounded-full pl-2 shadow-sm",
+        "ml-4 transition duration-200 ease-in-out hover:hover:bg-white/80",
+        "bg-white text-black"
+      )}
+    >
+      {children}
+    </Button>
+  );
+}
```
asim-shrestha, line 24: We should be reusing the same Button as the hero text. Why do we need to make custom ones?

next/src/components/NavLayout.tsx
```diff
@@ -45,11 +50,27 @@ export default function NavLayout({ children }: { children: ReactNode }) {
                       </a>
                     ))}
                   </div>
-                  <div className="hidden sm:ml-6 sm:flex sm:items-center">
-                    {/* Profile dropdown */}
-                    <Menu as="div" className="relative ml-3">
-                      <div></div>
-                    </Menu>
+                  <div className="ml-auto hidden sm:ml-6 sm:flex sm:items-center">
```
asim-shrestha, line 53: Why do we need the `ml-auto`?

PR #1036, opened 2023-07-13
🤫 Auth Redirects

awtkns:


next/src/hooks/useAuth.ts
```diff
@@ -14,23 +11,18 @@ interface Auth {
...
 
   useEffect(() => {
-    if (status != "authenticated" || !session?.user) return;
-
-    const user = session.user;
-    z.string()
-      .uuid()
-      .parseAsync(user.email)
-      .then(uuid => window.localStorage.setItem(UUID_KEY, uuid))
-      .catch(() => undefined);
```
asim-shrestha, line 28: what was this for?

asim-shrestha (approved):

next/src/hooks/useAuth.ts
awtkns: I am not sure... it might have been ported code

PR #1039, opened 2023-07-13
Fix port conflict between platform container and weaviate container

KennyDizi:
- Change weaviate container port to 8088 instead of 8080
- Update MySQL container version to v8.0.33
- Update Weaviate container version to v1.20.1

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,7 +37,7 @@ services:
       - db
 
   db:
-    image: mysql:8.0
+    image: mysql:8.0.33
     container_name: db
     restart: always
     build:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
       - db
 
   db:
-    image: mysql:8.0
+    image: mysql:8.0.33
```
awtkns, line 40: Would be best to just keep this a mysql 8 so that we are always using the latest version of mysql8

KennyDizi: That's a good idea, so let's keep it as v8.0

PR #1044, opened 2023-07-14
🔥 Ability to update input values on the Front end

asim-shrestha:
- Only deal with strings for now
- Nodes are created with empty strings for values
- Ability to update the values on the frontend
- The sidebar re-rendering is a bit jank ATM. Need to fix it

### DIFF
--- a/next/src/components/drawer/WorkflowSidebar.tsx
+++ b/next/src/components/drawer/WorkflowSidebar.tsx
@@ -66,6 +66,12 @@ const InspectSection = ({ selectedNode, updateNode }: InspectSectionProps) => {
 
   const definition = getNodeBlockDefinitions().find((d) => d.type === selectedNode.data.block.type);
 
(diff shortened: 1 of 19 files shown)


next/src/components/drawer/WorkflowSidebar.tsx
```diff
@@ -107,9 +119,14 @@ const NodeBlock = ({ definition, createNode }: NodeBlockProps) => {
   return (
     <div
       className="flex cursor-pointer flex-row gap-2 rounded-md border border-white/20 p-2 hover:bg-white/10"
-      onClick={() =>
-        createNode({ input: { url: "www.ThisIsARandomTestUrl.com" }, type: definition.type })
-      }
+      onClick={() => {
+        const input: Record<string, string> = {};
+        for (const field of definition.input_fields) {
```
awtkns, line 124: could do a map here

awtkns (approved):

PR #1048, opened 2023-07-14
🌟 Primary Button Glow

jasangill1:
## About 

- Added button glow, along with a hover effect that just make the opacity higher the duration was set to 1000 i think it look spretty cool
- Please let mw know on the use of the colours and hover animations

image


next/src/components/PrimaryButton.tsx
```diff
@@ -10,16 +10,19 @@ type PrimaryButtonProps = {
...
-        "rounded-full border-[1px] border-black shadow-sm",
-        "transition duration-200 ease-in-out hover:hover:bg-white/90 focus-visible:bg-white/90 focus-visible:outline-none focus-visible:ring-4 focus-visible:ring-white/30",
-        "bg-white text-black"
-      )}
-    >
-      {icon}
-      {children}
-    </Button>
+    <div className="group relative inline-flex items-center justify-center">
+      <div className="absolute -inset-1 rounded-full bg-gradient-to-r from-purple-700 to-purple-700 opacity-70 blur-lg transition-all duration-1000 group-hover:-inset-px group-hover:opacity-100 group-hover:duration-200" />
```
asim-shrestha: Lets extract this to a `Glow` component and just wrap the buttons with them

next/src/components/GlowWrapper.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+type GlowWrapperProps = {
+  children: React.ReactNode;
```
asim-shrestha, line 4: Can import `ReactNode` directly

next/src/components/GlowWrapper.tsx
```diff
@@ -0,0 +1,16 @@
...
+};
+
+const GlowWrapper: React.FC<GlowWrapperProps> = ({ children }) => {
```
asim-shrestha, line 7:
```suggestion
const GlowWrapper = ({ children }: GlowWrapperProps) => {
```

next/src/components/PrimaryButton.tsx
```diff
@@ -2,24 +2,28 @@ import clsx from "clsx";
...
-      className={clsx(
-        "rounded-full border-[1px] border-black shadow-sm",
-        "transition duration-200 ease-in-out hover:hover:bg-white/90 focus-visible:bg-white/90 focus-visible:outline-none focus-visible:ring-4 focus-visible:ring-white/30",
-        "bg-white text-black"
-      )}
-    >
-      {icon}
-      {children}
-    </Button>
+    <GlowWrapper>
```
asim-shrestha, line 15: We should wrap `PrimaryButton` wherever we want this glow wrapper to be

asim-shrestha: Not wrap it inside Primary Button itself

PR #1052, opened 2023-07-14
🔥 Display block outputs in the frontend

asim-shrestha:
- The frontend block definitions will now have output fields
- We will display those fields in a drop down when a node is selected
- We figure out fields by doing a reverse BFS from the current node up the tree

### DIFF
--- a/next/src/components/WindowButton.tsx
+++ b/next/src/components/WindowButton.tsx
@@ -14,7 +14,7 @@ const WindowButton = ({ ping, onClick, icon, name, border }: WindowButtonProps) 
   return (
     <div
       className={clsx(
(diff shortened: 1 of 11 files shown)


next/src/components/WindowButton.tsx
```diff
@@ -6,22 +6,22 @@ type WindowButtonProps = {
   ping?: boolean; // Toggles the ping animation
   onClick?: () => void;
   icon: React.ReactNode;
-  name: string;
+  text: string;
```
asim-shrestha, line 9: Had to change because apparently the "Name" is used as a key prop at times

awtkns (approved):

PR #1058, opened 2023-07-15
Update Dockerfile

sunzongjian:
update port


asim-shrestha:
Hey @sunzongjian, we'll also need to update the port on docker-compose to support this

PR #1063, opened 2023-07-16
Fix-1035: add link to status page in Error

abhithory:
Hello, @jasangill1 @asim-shrestha 

- I have added this link to the status page.
- this will show whenever any error will occur.
- if there are any changes in this. then please let me know.

image

### DIFF
--- a/next/src/components/console/ChatMessage.tsx
+++ b/next/src/components/console/ChatMessage.tsx
@@ -14,7 +14,7 @@ import {
   TASK_STATUS_STARTED,
 } from "../../types/task";
 import { FaCheck } from "react-icons/fa";
-import { FiClipboard } from "react-icons/fi";
+import { FiClipboard, FiExternalLink } from "react-icons/fi";
 
 const ChatMessage = ({ message }: { message: Message }) => {
   const [t] = useTranslation();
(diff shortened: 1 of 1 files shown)


next/src/components/console/ChatMessage.tsx
```diff
@@ -87,6 +90,17 @@ const FAQ = () => {
...
+    <p className="flex items-center gap-1">
+      Check Agent GPT Status{" "}
+      <a href="https://status.reworkd.ai" target="_blank" className="text-sky-500">
+       <FiExternalLink />
+      </a>
+    </p>
+  );
+};
+
 // Returns the translation key of the prefix
```
jasangill1, line 104: Could we indent so that the second line is starting at the same spot in the first line ?

jasangill1: Also no space between Agent GPT should be AgentGPT

abhithory: ok. i will do

abhithory:
hi @jasangill1 , done. Please check now

image

jasangill1:
indent should be aligned the as the ones on top

abhithory:
check this
image

jasangill1:
Looks good what do you think @asim-shrestha !!

PR #1096, opened 2023-07-19
🤖 openai + textinput blocks / logic added

shahrishabh7:


platform/reworkd_platform/schemas/workflow/blocks/openai_webhook.py
```diff
@@ -0,0 +1,49 @@
...
+        try:
+            response = create_llm(self.input.prompt)
+            logger.info(f"RESPONSE {response}")
+        except Exception as err:
+            logger.error(f"Failed to extract text with OpenAI: {err}")
+            raise
+
+        return OpenAIWebhookOutput(**self.input.dict(),result=response)
+    
+def create_llm(prompt):
```
awtkns, line 35: lets do this async if possible (i think you can use chain.arun)

platform/reworkd_platform/services/worker/exec.py
```diff
@@ -27,7 +27,7 @@ async def loop(self) -> None:
             self.workflow.workflow_id,
             "my-event",
             {
-                "nodeId": curr.id,
+                "nodeId": curr.id,
```
awtkns, line 30:
```suggestion
                "nodeId": curr.id,
```

awtkns (approved):

Issue #994, opened 2023-07-08
😵‍💫 Initial page load light mode and columns glitch


jasangill1:
As a user on initial load the flash of the light mode navbar and columns is distracting and takes away from user experience.

https://github.com/reworkd/AgentGPT/assets/56135840/74c32945-2ec5-4021-b44e-0e620bbedc4a

abhithory:
hi @jasangill1, I tried but this issue is not showing in my local development. 

- is this not showing in local development or is it not showing only in my pc?

jasangill1:
hi @abhithory could you try production and see if the issue is there?

abhithory:
@jasangill1 , yes, it is showing when I open the website from URL. but when I visit website locally (in both development and production mode) then it is not showing glitch and light mode

jasangill1:
Hello @abhithory Well that's apart of the problem figuring out what is causing this, comparing the other components whit the theme and seeing why they arent doing this on the initial load, troubleshooting and try to find the source of the problem

abhithory:
@jasangill1  ok, got it. let me try

kevinlu1248:
Hey I got Sweep (https://github.com/sweepai/sweep), an AI junior dev, to make a potential fix at https://github.com/kevinlu1248/AgentGPT/pull/8. Feel free to leave a comment on the PR or in the code prefixed with "Sweep:" and it will automatically get addressed by Sweep.

I also made a PR to this repo at https://github.com/reworkd/AgentGPT/pull/1099.

kevinlu1248:
Hey made another one at https://github.com/reworkd/AgentGPT/pull/1116

PR #1099, opened 2023-07-20
Flash of light mode fix

kevinlu1248:
This PR is copied from https://github.com/kevinlu1248/AgentGPT/pull/8, made by Sweep, an AI junior developer.

### What does this PR do?
This PR fixes the issue of a flash of light mode navbar and columns on the initial page load, which was distracting and took away from the user experience.

### How did you implement this feature/fix?
I modified the code in the Next.js application to ensure that the user's preferred theme is loaded immediately on initial page load, without defaulting to light mode first. Specifically, I made changes to the following files:

- next/src/pages/_app.tsx: Checked how the theme was being applied on initial page load and modified the code to apply the user's preferred theme immediately.
- next/src/pages/index.tsx: Checked how the theme was being applied on the home page and modified the code to apply the user's preferred theme immediately.

Fixes https://github.com/reworkd/AgentGPT/issues/994

### DIFF
--- a/next/src/pages/_app.tsx
+++ b/next/src/pages/_app.tsx
@@ -22,6 +22,15 @@ const MyApp: AppType<{ session: Session | null }> = ({
     document.documentElement.lang = i18n.language;
   }, [i18n]);
 
+  const preferredTheme = localStorage.getItem('theme');
+  useEffect(() => {
+    if (preferredTheme) {
+      document.documentElement.setAttribute('data-theme', preferredTheme);
+    } else {
+      document.documentElement.setAttribute('data-theme', 'light');
+    }
+  }, [preferredTheme]);
+
   return (
     <SessionProvider session={session}>
       <GoogleAnalytics trackPageViews />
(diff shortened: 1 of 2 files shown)


asim-shrestha:
Very cool @kevinlu1248! Unfortunately since we're using next js, this causes the build to fail as localstorage isn't available during SSR. Is this something Sweep can still tackle?

kevinlu1248:
@asim-shrestha I gave it a shot by prompting Sweep at https://github.com/kevinlu1248/AgentGPT/pull/8#issuecomment-1643364109 and looks like it's trying to use cookies, which I believe will also fail the SSR build. 

Frankly, I'm also not very familiar with how SSR, is the only way of solving this by setting the dark mode client-side? Would this even avoid the flash?

asim-shrestha:
@kevinlu1248 I found this: https://stackoverflow.com/questions/71277655/prevent-page-flash-in-next-js-12-with-tailwind-css-class-based-dark-mode

kevinlu1248:
Hey @asim-shrestha, made a new PR at https://github.com/reworkd/AgentGPT/pull/1116 based on that SO link

jasangill1:
fixed

PR #1102, opened 2023-07-20
🔥 Improve Glow effect

awtkns:
- noticed the glow was missing from the top right CTA
- noticed the glow should be a gradient from purple to blue instead of straight purple (subtle)
- tinkered with opacity to get it closer to figma look

### DIFF
--- a/next/src/components/GlowWrapper.tsx
+++ b/next/src/components/GlowWrapper.tsx
@@ -1,13 +1,20 @@
-import React from "react";
+import type { ReactNode } from "react";
+import clsx from "clsx";
 
 type GlowWrapperProps = {
-  children: React.ReactNode;
+  children: ReactNode;
+  className?: string;
 };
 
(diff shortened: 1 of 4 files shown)


shahrishabh7 (approved):

next/src/components/GlowWrapper.tsx
```diff
@@ -1,13 +1,20 @@
...
 };
 
-const GlowWrapper = ({ children }: GlowWrapperProps) => {
+const GlowWrapper = ({ children, className }: GlowWrapperProps) => {
   return (
     <div className="relative inline-flex items-center justify-center">
-      <div className="absolute -inset-1 rounded-full bg-[#A02BFE] opacity-50 blur-lg transition-all duration-1000" />
+      <div
+        className={clsx(
+          className || "opacity-50",
```
asim-shrestha, line 14: nit: might be better to call this `opacityClassName`

PR #1105, opened 2023-07-20
✨ Backend for IfNode

asim-shrestha:
- Store source_handle information in the backend
- An edge is only a duplicate if the source, handle, and source_handle are the same

- Also fixing an issue with the sidebar where it overflows


next/prisma/schema.prisma
```diff
@@ -149,14 +149,15 @@ model Workflow {
 }
 
 model WorkflowEdge {
-  id          String @id @default(cuid())
-  workflow_id String
-  source      String @db.VarChar(36)
-  target      String @db.VarChar(36)
+  id            String  @id @default(cuid())
```
awtkns, line 152: can you submit a planet scale pr for this?

awtkns (approved):

next/prisma/schema.prisma
asim-shrestha: Done!

PR #1107, opened 2023-07-20
🛠 Update DB CLI

jasangill1:


cli/src/envGenerator.js
```diff
@@ -13,14 +13,15 @@ export const generateEnv = (envValues) => {
         envValues,
         isDockerCompose,
         dbPort,
-        platformUrl
+        platformUrl,
+        isDockerCompose ? "/etc/ssl/certs/ca-certificates.crt" : ""
```
asim-shrestha, line 17: Already passing isDockerCompose into the `getEnvDefinition` function. 

`isDockerCompose ? "/etc/ssl/certs/ca-certificates.crt" : ""` This code should be use directly in the function where the path is defined

PR #1108, opened 2023-07-20
📂[WIP] File upload

awtkns:


platform/reworkd_platform/web/api/workflow/views.py
```diff
@@ -46,13 +47,17 @@ async def update_workflow(
     workflow_id: str,
     workflow: WorkflowUpdate,
     crud: WorkflowCRUD = Depends(WorkflowCRUD.inject),
-) -> str:
+) -> PresignedPost:
```
asim-shrestha, line 50: Should we decouple files from workflows a bit more? Maybe make a different route for uploads for example- feels a bit strange to stuff the post url here

next/src/pages/workflow/[workflow].tsx
```diff
@@ -13,13 +14,12 @@ import { useWorkflow } from "../../hooks/useWorkflow";
...
 import { languages } from "../../utils/languages";
 
-
-
-
 const WorkflowPage: NextPage = () => {
   const { session } = useAuth({ protectedRoute: true });
   const router = useRouter();
 
+  const [file, setFile] = useState<File>();
```
Comment (asim-shrestha) [inline next/src/pages/workflow/[workflow].tsx:21]: Should this be a part of useWorkflow? A bit strange to have everything else coupled to the workflow but have this state outside of it

asim-shrestha (approved):

asim-shrestha:
pytest failing

PR #1109, opened 2023-07-21
✏️ Workflow pdf extract

shahrishabh7:


platform/reworkd_platform/schemas/workflow/blocks/summarization_webhook.py
```diff
@@ -35,22 +46,51 @@ async def run(self) -> BlockIOBase:
...
         UserBase(id="", name=None, email="test@example.com"),
         streaming=False,
     )
     template = """
     You are a chatbot assistant that assists users in summarizing and extracting information from given text.
 
-    Question: {prompt}
+    Question: {prompt}. Here is the text: {text}
 
     Answer:"""
```
awtkns, line 90: We may need to iterate on this prompt a bit. Lets put this in our prompts.py file. There is also an example of our summary prompt for agent gpt in there

@asim-shrestha thoughts on the prompt?

platform/reworkd_platform/schemas/workflow/blocks/summarization_webhook.py
```diff
@@ -35,22 +46,51 @@ async def run(self) -> BlockIOBase:
...
+    return bytesIO_file
+
+
+def convert_pdf_to_string(bytesIO_file: BytesIO) -> str:
+    pdf_reader = pypdf.PdfReader(bytesIO_file)
+    extracted_text = ""
+
+    # Iterate over each page in the PDF
+    for page_num in range(len(pdf_reader.pages)):
+        # Get the current page
```
awtkns, line 67: You can do something like:
```suggestion
    for page in pdf_reader.pages:
        ...
```

platform/reworkd_platform/schemas/workflow/blocks/summarization_webhook.py
```diff
@@ -35,22 +46,51 @@ async def run(self) -> BlockIOBase:
...
+def fetch_file(filename: str) -> BytesIO:
+    session = boto3.Session(profile_name="dev")
+    REGION = "us-east-1"
+    bucket_name = "test-pdf-123"
+    s3_client = session.client("s3", region_name=REGION)
+    response = s3_client.get_object(Bucket=bucket_name, Key=filename)
+    response_body = response["Body"].read()
+    bytesIO_file = BytesIO(response_body)
+
+    return bytesIO_file
```
awtkns, line 58: Once my PR is merged we should put this in the S3 service

platform/reworkd_platform/schemas/workflow/blocks/summarization_webhook.py
```diff
@@ -35,22 +46,51 @@ async def run(self) -> BlockIOBase:
...
+
+
+def convert_pdf_to_string(bytesIO_file: BytesIO) -> str:
+    pdf_reader = pypdf.PdfReader(bytesIO_file)
+    extracted_text = ""
+
+    # Iterate over each page in the PDF
```
awtkns, line 65: Not sure comments are really needed here, the code already seems pretty semantically expressive. General rule of thumb i think is if you think you need a comment, it likely means you need to extract some logic into a function.

I like to think about the function name being a free comment of sorts :)

PR #1116, opened 2023-07-22
Fix flash of light mode on initial load

kevinlu1248:
Copied from https://github.com/kevinlu1248/AgentGPT/pull/13, by [Sweep](https://github.com/sweepai/sweep), an AI junior dev.

### Description
This PR addresses the issue of a brief flash of light mode during the initial load of the application, which is distracting for the user. The proposed solution is to create a JavaScript file that checks the user's preferred theme from local storage and applies it before the page loads. This file will be loaded using Next.js's Script component with a "beforeInteractive" strategy to ensure it runs before any other JavaScript. This approach should eliminate the flash by setting the correct theme before the page becomes interactive.

### Changes Made
- Created a new JavaScript file, `theme.js`, in the `public` directory to handle the theme setting before the page loads.
- Modified the `_app.tsx` file to include the `theme.js` file using the Next.js Script component with a "beforeInteractive" strategy.

Fixes https://github.com/reworkd/AgentGPT/issues/994

### DIFF
--- a/next/public/theme.js
+++ b/next/public/theme.js
@@ -1,0 +1,6 @@
+;(function initTheme() {
+  var theme = localStorage.getItem('theme') || 'light'
+  if (theme === 'dark') {
+    document.querySelector('html').classList.add('dark')
+  }
+})()
--- a/next/public/theme.js
+++ b/next/public/theme.js
@@ -1,6 +1,10 @@
 ;(function initTheme() {
-  var theme = localStorage.getItem('theme') || 'light'
-  if (theme === 'dark') {
-    document.querySelector('html').classList.add('dark')
-  }
+    var theme = localStorage.getItem('theme') || 'light'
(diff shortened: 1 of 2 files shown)


kevinlu1248:
Based on the preview it seems mostly fixed and just the sides broken.

asim-shrestha:
Hey @kevinlu1248 can you describe what this fixes? It appears the theme configuration is still not being correctly set (based on the sidebar light flash)

kevinlu1248:
@asim-shrestha It bundles the script for automatically setting the correct theme colour to the top of the HTML ensuring it gets ran earlier before everything else is hydrated. I'll see if Sweep can do the sides as well.

kevinlu1248:
Got Sweep to write a potential solution just now after looking through the dom tree for a bit

jasangill1:
fixed

PR #1117, opened 2023-07-22
🐛 Improve ssl DevX

awtkns:


asim-shrestha:
Need to remove the line in ENV generator that adds this

platform/reworkd_platform/services/ssl.py
```diff
@@ -0,0 +1,19 @@
+from ssl import create_default_context, SSLContext
+
+from reworkd_platform.settings import Settings
+
+MACOS_CERT_PATH = "/etc/ssl/cert.pem"
```
asim-shrestha, line 5: What if the user isn't on Mac?

awtkns: then their fucked

PR #1118, opened 2023-07-23
🐛 Fix for if condition

asim-shrestha:
- Add basic web interaction agent block with types
- Fix if condition to correctly prune branch

### DIFF
--- a/next/src/services/workflow/node-block-definitions.ts
+++ b/next/src/services/workflow/node-block-definitions.ts
@@ -6,7 +6,8 @@ import type { WorkflowNode } from "../../types/workflow";
 const IOFieldSchema = z.object({
   name: z.string(),
   description: z.string(),
-  type: z.enum(["string"]),
+  type: z.enum(["string", "array"]),
+  items: z.object({ type: z.string() }).optional(),
 });
 
 export type IOField = z.infer<typeof IOFieldSchema>;
@@ -78,13 +79,13 @@ const SummaryWebhookBlockDefinition: NodeBlockDefinition = {
     {
       name: "prompt",
       description: "What do you want to do with the text?",
-      type: "string"
+      type: "string",
     },
     {
       name: "filename",
       description: "reference a file that you want to summarize",
-      type: "string"
-    }
+      type: "string",
+    },
   ],
   output_fields: [
     {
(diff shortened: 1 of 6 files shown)


platform/reworkd_platform/services/worker/execution_engine.py
```diff
@@ -90,24 +93,38 @@ def create_execution_plan(
             ),
         )
 
-    def get_pruned_queue(self, if_condition: IfCondition, branch: bool) -> List[Node]:
+    def bfs_keep(self, node_ids_to_keep: List[str]) -> List[Node]:
+        visited = set(node_ids_to_keep)  # Initialize visited set with nodes to keep
+        queue = deque(node_ids_to_keep)  # Use a deque for efficient popleft operation
```
awtkns, line 98: Instead of leaving inline docs here. Might be good too leave a quick doc on the function eg:
```suggestion
    def bfs_keep(self, node_ids_to_keep: List[str]) -> List[Node]:
        """ explain algo """
        visited = set(node_ids_to_keep)
        queue = deque(node_ids_to_keep)
```

awtkns (approved):

next/src/services/workflow/node-block-definitions.ts
```diff
@@ -134,11 +140,34 @@ const TriggerBlockDefinition: NodeBlockDefinition = {
...
+  image_url: "/tools/web.png",
+  input_fields: [
+    {
+      name: "URL",
+      description: "The website the agent will interact with",
+      type: "string",
+    },
+    {
+      name: "Requirements",
+      description: "The steps the agent should take",
```
shahrishabh7, line 157: Maybe change name to 'Goal' ? Makes it clear that user can enter steps for agent to fetch specific information / execute actions.

shahrishabh7 (approved):

next/src/services/workflow/node-block-definitions.ts
asim-shrestha: Updated 🫡

platform/reworkd_platform/services/worker/execution_engine.py
asim-shrestha: Going to just remove comments actually. BFS is pretty standard

PR #1127, opened 2023-07-23
🐛Fix org sharing bugs

awtkns:


platform/reworkd_platform/db/crud/workflow.py
```diff
@@ -148,4 +149,4 @@ def validate_permissions(self, workflow: WorkflowModel) -> None:
         ):
             return
 
-        raise Exception("User does not have permission to update this workflow")
+        raise forbidden()
```
asim-shrestha, line 152: Missing string message

asim-shrestha (approved):

PR #1131, opened 2023-07-24
🚀Realtime features

awtkns:
- Presence when to users are on the same workflow (this is needed so we can reload on another users save)
- Socket server now uses authentication


asim-shrestha:
Lets fix mypy errors, goood to merge after

PR #1143, opened 2023-07-26
🐛 Fix rerender problem

awtkns:


next/src/layout/dashboard.tsx
```diff
@@ -72,39 +74,20 @@ const DashboardLayout = (props: DashboardLayoutProps) => {
...
-              setShow={setDesktop(rightSettings, setRightSettings)}
-            />
-          </div>
-          <div className={rightSettings.desktop ? "hidden" : "hidden lg:block"}>
-            <SidebarControlButton
-              side="right"
-              show={rightSettings.desktop}
-              setShow={setDesktop(rightSettings, setRightSettings)}
-            />
-          </div>
```
asim-shrestha, line 99: Any way to keep this? We don't want this always open by default for mobile as consumer AgentGPT will be affected

asim-shrestha: As a hack, we can make it hidden by default

awtkns: ahh yeah

PR #1146, opened 2023-07-26
🥥 Content Refresher Agent

KhoomeiK:
Added an agent that accepts a URL for an article and returns the content of the article along with a version of the article that has been refreshed from additional relevant sources found on the web


platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -0,0 +1,185 @@
...
+    print("MAIN")
+    agent = ContentRefresherAgent(
+        id="test",
+        type="ContentRefresherAgent",
+        input=ContentRefresherInput(
+            url="https://www.science.org/content/article/embattled-physicist-files-patent-unprecedented-ambient-superconductor"
+        )
+    )
+    output = agent.run()
+    print(output)
```
asim-shrestha, line 185: Lets delete this stuff, not relevant outside of local testing

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+    response_message = response.completion.strip()
+    new_info = "\n".join(response_message.split("\n\n"))
+    return new_info
+
+
+def add_info(target: str, info: str) -> str:
+    # Claude: rewrite target to include the info
+    prompt = f"Below are notes from some SOURCE articles:\n{info}\n----------------\nBelow is the TARGET article:\n{target}\n----------------\nPlease rewrite the TARGET article to include the information from the SOURCE articles."
+    response = claude.completions.create(
```
asim-shrestha, line 165: Might be worth it to create some abstraction over claude models or even just use langchain so we have retries with exponentail backoff, etc for prod

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -0,0 +1,172 @@
...
+    # use SERP API
+    response = requests.post(
+        f"https://google.serper.dev/search",
+        headers={
+            "X-API-KEY": settings.serp_api_key or '',
+            "Content-Type": "application/json",
+        },
+        params={
+            "q": search_query,
+        },
```
asim-shrestha, line 140: Should wrap this into a client abstraction and reuse across the app. Can be a separate ticket

PR #1147, opened 2023-07-27
📝 Change PDF to Docx output

asim-shrestha:


shahrishabh7 (approved):

platform/reworkd_platform/services/aws/s3.py
```diff
@@ -30,6 +32,33 @@ def upload_url(
...
+            Params={"Bucket": bucket_name, "Key": object_name},
+        )
+
+    async def upload_to_bucket(
+        self,
+        bucket_name: str,
+        object_name: str,
+        file: io.BytesIO,
+    ) -> None:
+        pre_signed_post = self.upload_url(bucket_name, object_name)
```
awtkns, line 47: You don't need a presigned post here. You can just upload directly using boto3 since this is server side

platform/pyproject.toml
```diff
@@ -39,13 +39,14 @@ pinecone-client = { version = "^2.2.2", extras = ["grpc"] }
 networkx = "^3.1"
 pusher = "^3.3.2"
 python-multipart = "^0.0.6"
-pypdf = "^3.13.0"
```
awtkns, line 42: This is needed for the summary agent

awtkns (changes requested):

platform/pyproject.toml
asim-shrestha: Re added

platform/reworkd_platform/services/aws/s3.py
asim-shrestha: Changed

PR #1154, opened 2023-07-27
🦙 Llama Ingestion class added + Google Sheets Input Block

shahrishabh7:
- Google Sheets Input block (writes to doc store)
- LlamaIndex connectors class created - wrote GoogleSheets input


next/src/services/workflow/node-block-definitions.ts
```diff
@@ -311,6 +311,27 @@ const FileUploadBlockDefinition: NodeBlockDefinition = {
...
 
+const GoogleSheetsBlockDefinition: NodeBlockDefinition = {
+  name: "Google Sheets Upload",
+  type: "GoogleSheetsUpload",
+  description: "Upload Google Sheets",
+  image_url: "/tools/web.png",
+  icon: FaBook,
+  input_fields: [
+    {
+      name: "sheet1_url",
```
asim-shrestha, line 322: `sheet_1_url`?

platform/reworkd_platform/schemas/workflow/blocks/google_sheets_upload.py
```diff
@@ -0,0 +1,81 @@
...
+    sheet2_url: str
+
+
+class GoogleSheetsUploadOutput(GoogleSheetsUploadInput):
+    None
+
+
+class GoogleSheetsUpload(Block):
+    type = "GoogleSheetsUpload"
+    description = "Upload Google Sheets"
```
asim-shrestha, line 38: Where is this uploading to? What are we uploading exactly?

platform/reworkd_platform/services/llama/llama.py
```diff
@@ -0,0 +1,97 @@
...
+                        flattened_string += cell["formattedValue"] + " "
+                flattened_string += "\n"
+
+        return flattened_string
+
+
+def test_sheets_reader():
+    # Replace 'YOUR_CREDENTIALS_PATH' with the path to your Google API credentials JSON file
+    creds_path = "YOUR_CREDENTIALS_PATH"
+    spreadsheet_id = "YOUR_SPREADSHEET_ID"
```
asim-shrestha, line 85: Should this be replaced? Also why is there a test function here?

platform/reworkd_platform/schemas/workflow/blocks/google_sheets_upload.py
asim-shrestha: Should this be called `GoogleSheetsLoader`?

PR #1157, opened 2023-07-28
🧠 Slack Oauth Flow

awtkns:


next/src/ui/OauthIntegration.tsx
awtkns, line 1: This component needs to be improved but it can be done after

shahrishabh7 (approved):

Issue #869, opened 2023-06-27
✨ Create a guide on how to update the documentation website


asim-shrestha:
We should update the docs site (https://docs.reworkd.ai/development) with some documentation for how to update the docs site itself. It should cover the following:

- Intro on what docusaurus is and link to it
- Talk about how to run the site locally
- Talk about how to update a page or create a new page
- Talk about exactly what we want documented
- Talk about how to request for documentation (Make an issue on GitHub!)

MusabShakeel576:
@asim-shrestha maybe we could update the [Docs README](https://github.com/reworkd/AgentGPT/blob/main/docs/README.md) since we already have a guide on how to run the docs site locally. Alternatively, we could update the [Docs Contributing](https://docs.reworkd.ai/contributing) file. Which option do you think would be better?

momokid:
@asim-shrestha  Hi, I would want to work on this issue.

asim-shrestha:
@momokid please go for it!

asim-shrestha:
Sorry I missed this @MusabShakeel576! I think a page on the docs site itself makes sense

MusabShakeel576:
@asim-shrestha no worries. Yeah that would be good.

jasangill1:
Hello @momokid Have you had any luck on this issue?? I can help you get started if not !!

momokid:
Hi @jasangill1 ,
I will be glad with all the help I can get. I was checking out muntlify, but I had some challenges. Sorry for not reverting sooner

jasangill1:
Hello @momokid what issues were u facing ??

momokid:
I wanted to run the mintlify setup locally and try out how it works, but I got an error during the installation. I was advised to install the previous mintlify version, but again, I got an error.

jasangill1:
are you using a new mintlify file? or are you running the AgentGPT docs??

momokid:
I am using the new mintlify file and I used the `mintlify dev` to start the mintlify server

momokid:
@jasangill1  This is the snapshot of the error
![mintlifydev](https://github.com/reworkd/AgentGPT/assets/23136074/d2b2c3c2-8d07-4da3-904f-678e8020157b)

momokid:
@jasangill1 or would you suggest I continue working on this issue without necessarily for this bug to be fixed?

jasangill1:
[...]

PR #1171, opened 2023-07-31
Guide on how to update the documentation website

momokid:
This is the new guide to updating the documentation website of the AgentGPT platform. It is in reference to the issue  #869


awtkns:
Hey thanks for this @momokid, what are your thoughts on putting this info under contributing?

momokid:
@awtkns I think it will be OK 👍

PR #1179, opened 2023-08-02
✨ Workflow QOL

asim-shrestha:
## About
Quality of line improvements when creating new nodes for the workflow. Essentially, we want to create the node where we clicked

## Notes
- Create node now takes in a `position` parameter for where to place the node once its created
- When you double click an area to create a node, it will store that click position and place the node there when it is added
- Edge is now animated even before save

### DIFF
--- a/next/src/components/workflow/Flowchart.tsx
+++ b/next/src/components/workflow/Flowchart.tsx
@@ -77,23 +77,24 @@ const FlowChart = forwardRef<FlowChartHandles, FlowChartProps>(
     );
 
     const onConnectStart = useCallback(() => {
-      setConnectionDragging(true);
-    }, [setConnectionDragging]);
+      connectionDragging.current = true;
+    }, [connectionDragging]);
 
     const onConnect = useCallback(
       (connection: Connection) => {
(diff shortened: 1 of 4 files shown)


next/src/components/workflow/Flowchart.tsx
```diff
@@ -75,13 +103,27 @@ const FlowChart = forwardRef<FlowChartHandles, FlowChartProps>(
...
 
+    const onConnectEnd = useCallback(
+      (event: MouseEvent | TouchEvent) => {
+        if (!connectionDragging.current) return;
+        connectionDragging.current = false;
+        onPaneDoubleClick(getExactPosition(event));
+      },
+      [getExactPosition, onPaneDoubleClick, connectionDragging]
+    );
+
```
asim-shrestha, line 126: Essentially, when we make an edge and connect it to nothing, we should pop open the creation dialog. In the future, when a node is created this way we should also attach the edge to it.

shahrishabh7: What do you mean by attach the edge to it?

PR #1180, opened 2023-08-03
👈 Remove pointer events

asim-shrestha:
- The nav section is un-clickable because of pointer events. A bit annoying innit?
- No select items available message. Otherwise it "opens" to almost nothing

image

### DIFF
--- a/next/src/pages/workflow/index.tsx
+++ b/next/src/pages/workflow/index.tsx
@@ -1,4 +1,5 @@
 import { useQuery } from "@tanstack/react-query";
+import clsx from "clsx";
 import type { GetStaticProps } from "next";
 import { type NextPage } from "next";
 import Image from "next/image";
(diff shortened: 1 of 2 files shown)


next/src/pages/workflow/index.tsx
```diff
@@ -244,7 +247,7 @@ const WorkflowPage: NextPage = () => {
       )}
 
       {!showLoader && !showCreateForm && !nodesModel[0].length && (
-        <div className="fixed left-1/2 top-1/2 z-10 -translate-x-1/2 -translate-y-1/2 rounded-md border-2 border-dashed border-black p-8 text-lg font-light tracking-wider backdrop-blur-[2px]">
+        <div className="pointer-events-none fixed left-1/2 top-1/2 z-10 -translate-x-1/2 -translate-y-1/2 rounded-md border-2 border-dashed border-black p-8 text-lg font-light tracking-wider backdrop-blur-[2px]">
```
asim-shrestha, line 250: This is the message that pops up when there is no nodes in the workflow. Instinctively, a user will double click it with nothing happening

PR #1184, opened 2023-08-03
🏎️ Make content refresher async

awtkns:


platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -41,20 +41,21 @@ async def run(self, workflow_id: str, **kwargs: Any) -> ContentRefresherOutput:
         logger.info(source_urls)
 
         source_contents = [
-            get_page_content(url)
+            await get_page_content(url)
```
awtkns, line 44: cant do this in parallel yet bc of rate limits

shahrishabh7 (approved):

PR #1185, opened 2023-08-03
🌎 Pass source metadata through content refresher agent

shahrishabh7:
- pass source metadata through content refresher agent (cite in-line on added information)
- modify `add_info()` prompt so result follows `target_content` format better

### DIFF
--- a/platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
+++ b/platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
@@ -134,7 +134,7 @@ async def find_content_kws(content: str) -> str:
     )
 
 
-def search_results(search_query: str) -> list[dict[str, str, str]]:
+def search_results(search_query: str) -> list[dict[str, str]]:
     # use SERP API
     response = requests.post(
         f"https://google.serper.dev/search",
@@ -158,8 +158,10 @@ def search_results(search_query: str) -> list[dict[str, str, str]]:
     return source_information
 
 
-async def find_new_info(target: str, source: dict[str, str, str]) -> str:
-    source_metadata = f"{source['url']}, {source['title']}, {source['date']}"
+async def find_new_info(target: str, source: dict[str, str]) -> str:
+    source_metadata = f"{source['url']}, {source['title']}" + (
+        f", {source['date']}" if source["date"] else ""
+    )
     source_content = source["content"]
 
     # Claude: info mentioned in source that is not mentioned in target
--- a/platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
+++ b/platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
@@ -156,12 +156,12 @@ def search_results(search_query: str) -> list[tuple[str, str]]:
 
 
 async def find_new_info(target: str, source: tuple[str, str, str]) -> str:
-    source_info = f"{source[0]}, {source[1]}"
+    source_metadata = f"{source[0]}, {source[1]}"
     source_content = source[2]
 
     # Claude: info mentioned in source that is not mentioned in target
     prompt = HumanAssistantPrompt(
-        human_prompt=f"Below is the TARGET article:\n{target}\n----------------\nBelow is the SOURCE article:\n{source_content}\n----------------\nIn a bullet point list, identify all facts, figures, or ideas that are mentioned in the SOURCE article but not in the TARGET article. Under the list, write a source citation (you MUST include the url AND full source title provided below: \n{source_info}",
+        human_prompt=f"Below is the TARGET article:\n{target}\n----------------\nBelow is the SOURCE article:\n{source_content}\n----------------\nIn a bullet point list, identify all facts, figures, or ideas that are mentioned in the SOURCE article but not in the TARGET article. Under the list, write a source citation (you MUST include the url AND full source title provided below: \n{source_metadata}",
         assistant_prompt="Here is a list of claims in the SOURCE that are not in the TARGET:",
     )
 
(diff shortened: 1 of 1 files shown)


platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -35,21 +35,23 @@ async def run(self, workflow_id: str, **kwargs: Any) -> ContentRefresherOutput:
...
+        sources = search_results(keywords)
+        sources = [
+            (url, title) for url, title in sources if url != target_url
+        ]  # TODO: check based on content overlap
+        logger.info(sources)
 
         source_contents = [
-            await get_page_content(url)
-            for url in source_urls[:3]  # TODO: remove limit of 3 sources
+            (url, title, await get_page_content(url)) for url, title in sources[:3]
```
KhoomeiK, line 45: Since this is async now, are we ready to remove the 3 source limit?

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -35,21 +35,23 @@ async def run(self, workflow_id: str, **kwargs: Any) -> ContentRefresherOutput:
...
         ]
 
         source_contents = [
-            content for content in source_contents if content is not None
+            (url, title, page_content)
+            for url, title, page_content in source_contents
+            if page_content is not None
```
KhoomeiK, line 51: Would be fairly straightforward to add a string similarity condition here to check if page_content is similar to target_content.

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -147,14 +149,19 @@ def search_results(search_query: str) -> list[str]:
         },
     )
     response.raise_for_status()
-    urls = [result["link"] for result in response.json()["organic"]]
-    return urls
+    source_information = [
+        (result["link"], result["title"]) for result in response.json()["organic"]
```
KhoomeiK, line 153: Can we also include the publication date from Serper API? Customer priority is novel information, so sorting by date would be useful.

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -147,14 +149,19 @@ def search_results(search_query: str) -> list[str]:
...
 
+async def find_new_info(target: str, source: tuple[str, str, str]) -> str:
+    source_metadata = f"{source[0]}, {source[1]}"
+    source_content = source[2]
 
-async def find_new_info(target: str, source: str) -> str:
     # Claude: info mentioned in source that is not mentioned in target
     prompt = HumanAssistantPrompt(
-        human_prompt=f"Below is the TARGET article:\n{target}\n----------------\nBelow is the SOURCE article:\n{source}\n----------------\nIn a bullet point list, identify all facts, figures, or ideas that are mentioned in the SOURCE article but not in the TARGET article.",
+        human_prompt=f"Below is the TARGET article:\n{target}\n----------------\nBelow is the SOURCE article:\n{source_content}\n----------------\nIn a bullet point list, identify all facts, figures, or ideas that are mentioned in the SOURCE article but not in the TARGET article. Under the list, write a source citation (you MUST include the url AND full source title provided below: \n{source_metadata}",
```
KhoomeiK, line 164: Why can't we just append `source_metadata` after we get the response? Like after declaring `new_info`, we can add `new_info += '\n' + source_metadata`.

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -170,7 +177,7 @@ async def find_new_info(target: str, source: str) -> str:
 async def add_info(target: str, info: str) -> str:
     # Claude: rewrite target to include the info
     prompt = HumanAssistantPrompt(
-        human_prompt=f"Below are notes from some SOURCE articles:\n{info}\n----------------\nBelow is the TARGET article:\n{target}\n----------------\nPlease rewrite the TARGET article to include the information from the SOURCE articles.",
+        human_prompt=f"Below are notes from some SOURCE articles:\n{info}\n----------------\nBelow is the TARGET article:\n{target}\n----------------\nPlease rewrite the TARGET article to include the information from the SOURCE articles. Maintain the format of the TARGET article. At the end of the article, include a list of sources (source url, title, and any additional information) ONLY for added information from SOURCE articles using the following example format: 'Source: https://www.wisnerbaum.com/prescription-drugs/gardasil-lawsuit/, Gardasil Vaccine Lawsuit Update August 2023 - Wisner Baum' Do not add citations for any info in the TARGET article.",
```
KhoomeiK, line 180: Is this the behavior we want or do we want inline citations? If we just want to append citations at the end, we can simply append to the string (no need to include in prompt to LLM). If the latter, this prompt will need to be edited slightly.

KhoomeiK (changes requested):
Looks great for the most part, main priorities imo are clarifying how we want to deal with citations (see comments) and potentially adding publication date info from Serper API.

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
shahrishabh7: fixed to in-line citations.

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
shahrishabh7: Serp payload doesn't contain publication dates

SERP payload (gotta scroll a lil): https://serpapi.com/search-api

Example of a source article: https://www.millerandzois.com/products-liability/gardasil/
Is there an alt way to get publication dates for content like this?

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
shahrishabh7: Yessir, I'll add this in another PR

awtkns (approved):

PR #1204, opened 2023-08-07
♟️Competitor check

shahrishabh7:
Content refresher agent - check / remove sources for refresh from competitors 
- added input to ContentRefresherAgent


platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -49,14 +50,16 @@ def log(msg: Any) -> None:
         log("Finding sources to refresh content")
         log("\n".join([f"- {source['title']}: {source['url']}" for source in sources]))
 
+        log("Checking new sources for competitors")
```
asim-shrestha, line 53: Maybe `Removing competitors from sources`

and then another log for removed sources?

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -169,6 +172,25 @@ def search_results(search_query: str) -> List[Dict[str, str]]:
     return source_information
 
 
+def remove_competitors(
+    sources: List[Dict[str, str]], competitors: List[str], log: Callable[[str], None]
+) -> List[Dict[str, str]]:
+    normalized_competitors = [comp.replace(" ", "").lower() for comp in competitors]
```
asim-shrestha, line 178: Can you use `.trim()` here for `normalization`?

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -169,6 +172,25 @@ def search_results(search_query: str) -> List[Dict[str, str]]:
...
+    filtered_sources = []
+    for source in sources:
+        if competitor_pattern.search(
+            source["url"].replace(" ", "").lower()
+        ) or competitor_pattern.search(source["title"].replace(" ", "").lower()):
+            log(f"Removing source due to competitor match:, '{source['title']}'")
+        else:
+            filtered_sources.append(source)
+
+    return filtered_sources
```
asim-shrestha, line 191: We should write a test for this

asim-shrestha (approved):

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
shahrishabh7: submitting in next PR.

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -169,6 +172,25 @@ def search_results(search_query: str) -> List[Dict[str, str]]:
     return source_information
 
 
+def remove_competitors(
+    sources: List[Dict[str, str]], competitors: List[str], log: Callable[[str], None]
+) -> List[Dict[str, str]]:
+    normalized_competitors = [comp.replace(" ", "").lower() for comp in competitors]
+    competitor_pattern = re.compile(
+        "|".join(re.escape(comp.lower()) for comp in normalized_competitors)
```
awtkns, line 180: probably don't need to lower it again here

awtkns: Some refactoring for clarity could be beneficial here:
```python3
from typing import List, Dict, Callable
import re

def normalize_string(value: str) -> str:
    return value.replace(" ", "").lower()

def create_competitor_pattern(competitors: List[str]) -> re.Pattern:
    normalized_competitors = [normalize_string(comp) for comp in competitors]
    pattern_string = "|".join(re.escape(comp) for comp in normalized_competitors)
    return re.compile(pattern_string)

def is_competitor_match(pattern: re.Pattern, source: Dict[str, str]) -> bool:
    for key in ["url", "title"]:
        if pattern.search(normalize_string(source[key])):
            return True
    return False

def remove_competitors(
    sources: List[Dict[str, str]], 
    competitors: List[str], 
    log: Callable[[str], None]
) -> List[Dict[str, str]]:
    
    competitor_pattern = create_competitor_pattern(competitors)
    filtered_sources = [
        source for source in sources
        if not (is_competitor_match(competitor_pattern, source) and log(f"Removing source due to competitor match:, '{source['title']}'"))
    ]

    return filtered_sources
```

PR #1207, opened 2023-08-07
✏️ Content agent bullet output

shahrishabh7:
- additional output on content refresher agent (bullet points of infos from each source)
- fix 'workflow failed: too much data' error (iterate and log new_infos rather than one-liner)
- make `competitors` input optional

### DIFF
--- a/platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
+++ b/platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
@@ -1,5 +1,5 @@
 import re
-from typing import Any, Callable, Dict, List
+from typing import Any, Callable, Dict, List, Optional
 
 import requests
 from bs4 import BeautifulSoup
@@ -14,7 +14,7 @@ from reworkd_platform.settings import settings
 
 class ContentRefresherInput(BlockIOBase):
     url: str
-    competitors: str
+    competitors: Optional[str] = None
 
 
 class ContentRefresherOutput(ContentRefresherInput):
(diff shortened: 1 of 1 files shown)


platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -14,12 +14,13 @@
...
 class ContentRefresherInput(BlockIOBase):
     url: str
-    competitors: str
+    competitors: Optional[str] = None
 
 
 class ContentRefresherOutput(ContentRefresherInput):
     original_content: str
-    refreshed_content: str
+    refreshed_report: str
```
asim-shrestha, line 22: Why the name change?

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -69,7 +70,8 @@ def log(msg: Any) -> None:
 
         new_infos = "\n\n".join(new_info)
         log("Extracting new, relevant information not present in your content")
-        log(new_infos)
+        for info in new_info:
+            log(info)
```
asim-shrestha, line 74: Should this be bullet points?

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -185,7 +188,7 @@ def remove_competitors(
         if competitor_pattern.search(
             source["url"].replace(" ", "").lower()
         ) or competitor_pattern.search(source["title"].replace(" ", "").lower()):
-            log(f"Removing source due to competitor match:, '{source['title']}'")
+            log(f"Removing source due to competitor match: '{source['title']}'")
```
asim-shrestha, line 191: Maybe `Removing competitive source:` is better wording

asim-shrestha (approved):

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
shahrishabh7: because I added parameter `refreshed_bullet_points` as well; `refreshed_content` feels too generic / like it could apply to either string body, figured `refreshed_report` / `refreshed_bullet_points` were better descriptors

platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
shahrishabh7: nope, they're already logging as bullet points (prompted in `add_info()`)

PR #1210, opened 2023-08-07
Content Refresher agent prompt update for get_page_content

KhoomeiK:
Also includes DB Connection Checker block


platform/reworkd_platform/web/api/workflow/blocks/web/__init__.py
```diff
@@ -36,6 +36,8 @@ def get_block_runner(block: Block) -> Block:
         return UploadDoc(**block.dict())
     if block.type == "DiffDoc":
         return DiffDoc(**block.dict())
+    if block.type == "DB":
+        return DiffDoc(**block.dict())
```
asim-shrestha, line 40: We should remove this. Was just for testing during the demo right?

next/src/services/workflow/node-block-definitions.ts
```diff
@@ -247,6 +247,34 @@ const DiffDocBlockDefinition: NodeBlockDefinition = {
...
+  ],
+  output_fields: [
+    {
+      name: "file_url",
+      description: "The URL to access the diff PDF.",
+      type: "string",
+    },
+  ],
+};
+
```
asim-shrestha, line 277: We should remove this. Was just for testing during the demo right?

PR #1217, opened 2023-08-09
🤖 Content refresher refresh

shahrishabh7:
fix for Claude line range parsing bug


platform/reworkd_platform/schemas/workflow/blocks/agents/content_refresher_agent.py
```diff
@@ -132,16 +131,26 @@ async def get_page_content(self, url: str) -> str:
         content = []
         for line_num in line_nums.split(","):
             if "-" in line_num:
-                start, end = map(int, line_num.split("-"))
-                for i in range(start, end + 1):
-                    text = ".".join(pgraphs[i - 1].split(".")[1:]).strip()
-                    content.append(text)
-            else:
+                start, end = self.extract_initial_line_numbers(line_num)
```
awtkns, line 134: we'll likely need to refactor this. Might be best to put it in a separate function in the future to give the reader a better understanding of what it is doing

awtkns (approved):

PR #1254, opened 2023-08-27
Generate personalized agent goals for personal data interaction

gvilums:
This PR introduces functionality to generate a custom agent goal when the user interacts with AssistantGPT. This goal is generated using the SID "example" endpoint, which generates prompts that could be answered with the user's connected data sources. The goal of this is to better show users the possible uses of a personal data agent, and to allow them to get a better out of the box experience.

Furthermore, the active tools used by the AssistantGPT example agent are modified to only include SID, allowing it to better stay on track and use the SID tool. I'm not too sure about this change: It definitely improves the agent's performance when dealing with personal data and the SID tool, but it may be confusing to the user.


gvilums:
Something seems to have gone wrong with rebasing, i'll get that figured out first.

PR #1304, opened 2023-09-30
fix: add httpx dependency to platform Dockerfile

JonasLaux:
- Add necessary dependency for backend


darkanubis0100 (approved):

awtkns:
@JonasLaux looks good but this should just be added to the pyproject.toml file like other dependacies

PR #1342, opened 2023-10-27
In SID tool, fall back to normal search if no results are available

gvilums:
These changes make it so that, if the SID tool cannot find any relevant information, or if the tool experiences some other issue (e.g. expired token, network issues, etc.), the agent automatically falls back to a standard google search.

Also, alignment of buttons in the tool menu is fixed for the SID tool.

### DIFF
--- a/platform/reworkd_platform/web/api/agent/tools/sidsearch.py
+++ b/platform/reworkd_platform/web/api/agent/tools/sidsearch.py
@@ -137,13 +137,7 @@ class SID(Tool):
         *args: Any,
         **kwargs: Any,
     ) -> FastAPIStreamingResponse:
-        res = await self._run_sid(
-            goal, task, input_str, user, oauth_crud
-        )
-        if res is None:
-            # fall back to search if no results are found
-            return await Search(self.model, self.language).call(
-                goal, task, input_str, user, oauth_crud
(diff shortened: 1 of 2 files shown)


platform/reworkd_platform/web/api/agent/tools/sidsearch.py
```diff
@@ -122,9 +122,31 @@ async def call(
...
+        res = await self._run_sid(
+            goal, task, input_str, user, oauth_crud
+        )
+        if res is None:
+            # fall back to search if no results are found
+            return await Search(self.model, self.language).call(
+                goal, task, input_str, user, oauth_crud
+            )
+        else:
+            return res
```
awtkns, line 152: Lets get ride of the else statement here. 
```suggestion
         # fall back to search if no results are found
        return await self._run_sid(goal, task, input_str, user, oauth_crud) or await Search(self.model, self.language).call(
        goal, task, input_str, user, oauth_crud
    )
```

awtkns (approved):

awtkns:
@gvilums would you be able to update this with main so the ci can run :)

gvilums:
@awtkns I commited the change you suggested and rebased onto main.

PR #1546, opened 2024-05-06
Update OpenAI API key regular expression 

jmrada14:
https://github.com/reworkd/AgentGPT/pull/219 introduces the previous regular expression but as mentioned in issue current keys have 56 chars instead of 51. removing the length validation in case it changes again in the future.


sidPhoenix17:
nobody is looking into this project?

PR #1666, opened 2025-04-05
Fix platform - poetry issue with newer versions of python packages

eufrat987:
Issues with latest versions: 
- No module named 'langchain.globals'
- pydantic_settings instead of pydantic
- validation: Extra inputs are not permitted

Imo Easiest fix


athuljayaram:
looks good, I authorize this deployment of the current commit.

